Issue #7, opened 2018-10-08
Provide capability to retrieve a secret from Azure Key Vault


tomkerkhove:
Provide capability to retrieve a secret.

In the first version, we should only support basic authentication.

We should provide an `AzureKeyVaultProvider` that implements `ISecretProvider` (#20) which is reading secrets from Azure Key Vault.

This should end up in the `Arcus.Security.AzureKeyVault` 📦 .

stijnmoreels:
This seems like the first required and useful feature. This and #22 can be assigned to me if you like. 

Maybe we have to have a chat on the specs or are they have to be specified in here (not sure how big this is, hence my question).

tomkerkhove:
We'll need to keep in mind that we will have different flavors of authentication so we should abstract that concept out via an interface or something related.

See #10 & #11

SamVanhoutte:
The Authentication should indeed make abstraction of the actual implementation (we can inject the right SecretProvider).  
I just wonder what's the best approach to configure the different key providers (KeyVault will have different config values than others, for example)

tomkerkhove:
Can you provide docs for this please @SamVanhoutte?

PR #35, opened 2018-12-18
Key vault support

SamVanhoutte:
PR to implement issue #7

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -328,3 +328,4 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
+/src/Arcus.Security.Tests.Integration/appsettings.json
--- a/src/Arcus.Security.KeyVault/Factories/KeyVaultClientFactory.cs
+++ b/src/Arcus.Security.KeyVault/Factories/KeyVaultClientFactory.cs
@@ -1,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using Microsoft.Azure.KeyVault;
+
+namespace Arcus.Security.KeyVault.Factories
+{
+    public abstract class KeyVaultClientFactory
+    {
+        public abstract Task<KeyVaultClient> CreateClient();
+    }
+}
--- a/src/Arcus.Security.KeyVault/Factories/KeyVaultClientFactory.cs
+++ b/src/Arcus.Security.KeyVault/Factories/KeyVaultClientFactory.cs
@@ -3,8 +3,15 @@ using Microsoft.Azure.KeyVault;
 
 namespace Arcus.Security.KeyVault.Factories
 {
+    /// <summary>
+    /// Abstract class to create <see cref="KeyVaultClient"/> object
+    /// </summary>
     public abstract class KeyVaultClientFactory
     {
+        /// <summary>
+        /// Method to implement the creation of your own KeyVaultClient
+        /// </summary>
+        /// <returns>An initialized KeyVaultClient</returns>
         public abstract Task<KeyVaultClient> CreateClient();
     }
 }
--- a/src/Arcus.Security.KeyVault/Factories/MsiKeyVaultClientFactory.cs
+++ b/src/Arcus.Security.KeyVault/Factories/MsiKeyVaultClientFactory.cs
@@ -5,8 +5,15 @@ using Microsoft.Azure.Services.AppAuthentication;
 
 namespace Arcus.Security.KeyVault.Factories
 {
+    /// <summary>
+    /// <see cref="KeyVaultClientFactory"/> implementation using Azure Managed Service Identity
+    /// </summary>
     public class MsiKeyVaultClientFactory : KeyVaultClientFactory
     {
+        /// <summary>
+        /// Creates a <see cref="KeyVaultClient"/>, using the AzureServiceTokenProvider
+        /// </summary>
+        /// <returns>A generated KeyVaultClient</returns>
         public override Task<KeyVaultClient> CreateClient()
         {
             var tokenProvider = new AzureServiceTokenProvider();
--- a/src/Arcus.Security.KeyVault/Factories/MsiKeyVaultClientFactory.cs
+++ b/src/Arcus.Security.KeyVault/Factories/MsiKeyVaultClientFactory.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.KeyVault;
+using Microsoft.Azure.Services.AppAuthentication;
+
+namespace Arcus.Security.KeyVault.Factories
+{
+    public class MsiKeyVaultClientFactory : KeyVaultClientFactory
+    {
+        public override Task<KeyVaultClient> CreateClient()
+        {
+            var tokenProvider = new AzureServiceTokenProvider();
+            return Task.FromResult(new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(tokenProvider.KeyVaultTokenCallback)));
+        }
+    }
+}
--- a/src/Arcus.Security.KeyVault/Factories/ServicePrincipalKeyVaultClientFactory.cs
+++ b/src/Arcus.Security.KeyVault/Factories/ServicePrincipalKeyVaultClientFactory.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Threading.Tasks;
+using GuardNet;
+using Microsoft.Azure.KeyVault;
+using Microsoft.IdentityModel.Clients.ActiveDirectory;
+
+namespace Arcus.Security.KeyVault.Factories
+{
+    public class ServicePrincipalKeyVaultClientFactory : KeyVaultClientFactory
+    {
+        private readonly string _clientId;
+        private readonly string _clientKey;
+
+        /// <summary>
+        /// Initializes <see cref="ServicePrincipalKeyVaultClientFactory"/> that will generate a KeyVaultClient, using a service principal
+        /// </summary>
+        /// <param name="clientId">The ClientId of the service principal, used to connect to Azure Key Vault</param>
+        /// <param name="clientKey">The Secret ClientKey of the service principal, used to connect to Azure Key Vault</param>
+        public ServicePrincipalKeyVaultClientFactory(string clientId, string clientKey)
+        {
+            Guard.NotNullOrEmpty(clientId, nameof(clientId));
+            Guard.NotNullOrEmpty(clientKey, nameof(clientKey));
+
+            _clientId = clientId;
+            _clientKey = clientKey;
+        }
+
+        public override Task<KeyVaultClient> CreateClient()
+        {
+            var keyVaultClient = new KeyVaultClient(GetToken);
+            return Task.FromResult(keyVaultClient);
+        }
+
+        private async Task<string> GetToken(string authority, string resource, string scope)
+        {
+            var authContext = new AuthenticationContext(authority);
+            ClientCredential clientCred = new ClientCredential(_clientId, _clientKey);
+            AuthenticationResult result = await authContext.AcquireTokenAsync(resource, clientCred);
+
+            if (result == null)
+            {
+                throw new InvalidOperationException("Failed to obtain the JWT token");
+            }
+
+            return result.AccessToken;
+        }
+    }
+}
--- a/src/Arcus.Security.KeyVault/Factories/ServicePrincipalKeyVaultClientFactory.cs
+++ b/src/Arcus.Security.KeyVault/Factories/ServicePrincipalKeyVaultClientFactory.cs
@@ -6,6 +6,9 @@ using Microsoft.IdentityModel.Clients.ActiveDirectory;
 
 namespace Arcus.Security.KeyVault.Factories
 {
+    /// <summary>
+    /// <see cref="KeyVaultClientFactory"/> implementation using Azure Managed Service Identity
+    /// </summary>        
     public class ServicePrincipalKeyVaultClientFactory : KeyVaultClientFactory
     {
         private readonly string _clientId;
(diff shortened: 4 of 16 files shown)


.gitignore
```diff
@@ -328,3 +328,4 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
+/src/Arcus.Security.Tests.Integration/appsettings.json
```
tomkerkhove: Feel free to align with our Event Grid approach and provide a tokenized config https://github.com/arcus-azure/arcus.eventgrid/blob/master/src/Arcus.EventGrid.Tests.Integration/appsettings.json

src/Arcus.Security.KeyVault/Factories/KeyVaultClientFactory.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Threading.Tasks;
+using Microsoft.Azure.KeyVault;
+
+namespace Arcus.Security.KeyVault.Factories
+{
+    /// <summary>
+    /// Abstract class to create <see cref="KeyVaultClient"/> object
+    /// </summary>
+    public abstract class KeyVaultClientFactory
```
tomkerkhove: Just curious - Why did you go for an abstract class over an interface?

src/Arcus.Security.KeyVault/Factories/MsiKeyVaultClientFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class MsiKeyVaultClientFactory : KeyVaultClientFactory
+    {
+        /// <summary>
+        /// Creates a <see cref="KeyVaultClient"/>, using the AzureServiceTokenProvider
+        /// </summary>
+        /// <returns>A generated KeyVaultClient</returns>
+        public override Task<KeyVaultClient> CreateClient()
+        {
+            var tokenProvider = new AzureServiceTokenProvider();
+            return Task.FromResult(new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(tokenProvider.KeyVaultTokenCallback)));
```
tomkerkhove: There's a lot of inlining here for new objects, a bit painful from a code readability standpoint.

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.Extensions.Configuration;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Arcus.Security.Tests.Integration.KeyVault
+{
+    [Trait(name: "Category", value: "Integration")]
+    public class KeyVaultSecretProviderTests
+    {
+        // The same tests should be tested with different KeyVaultClientFactories
```
tomkerkhove: I personally would create dedicated methods for those given they are testing different things.
The alternative is passing them via a test parameter.

I think for all tests we should pass them via test parameter, but have dedicated ones for happy path as well given you'd want to test the factory itself as well.

src/Arcus.Security.Tests.Integration/Logging/XunitTestLogger.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit.Abstractions;
+using Microsoft.Extensions.Logging;
+using GuardNet;
+
+namespace Arcus.Security.Tests.Integration.Logging
+{
+    public class XunitTestLogger : ILogger
```
tomkerkhove: Looks like copy paste from event grid, is that correct?

If so, please add an issue https://github.com/arcus-azure/arcus to create https://github.com/arcus-azure/arcus.core with something like this.

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -0,0 +1,47 @@
...
+using Arcus.Security.KeyVault;
+using Arcus.Security.KeyVault.Factories;
+using Microsoft.Azure.KeyVault;
+using Xunit;
+
+namespace Arcus.Security.Tests.Unit.KeyVault
+{
+    public class KeyVaultSecretProviderTests
+    {
+        [Fact]
```
tomkerkhove: We're testing different things here so should be separate tests imo

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -0,0 +1,76 @@
...
+using Arcus.Security.KeyVault.Factories;
+using Arcus.Security.Tests.Integration.Logging;
+using Microsoft.Extensions.Configuration;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Arcus.Security.Tests.Integration.KeyVault
+{
+    [Trait(name: "Category", value: "Integration")]
+    public class KeyVaultSecretProviderTests
```
tomkerkhove: Can you create the integration test build + ARM template for this as well please?
We can create an issue for it but then it will never happen/be postponed too long.

src/Arcus.Security.KeyVault/KeyVaultSecretProvider.cs
```diff
@@ -1,16 +1,69 @@
...
+        public KeyVaultSecretProvider(KeyVaultClientFactory keyVaultClientFactory, string keyVaultUri)
+        {
+            Guard.NotNullOrEmpty(keyVaultUri, nameof(keyVaultUri));
+            Guard.NotNull(keyVaultClientFactory, nameof(keyVaultClientFactory));
+            _keyVaultUri = keyVaultUri;
+            _keyVaultClientFactory = keyVaultClientFactory;
+        }
+
+        // Moving this implementation here, as KeyVaultClient cannot be mocked (https://github.com/Azure/azure-sdk-for-java/issues/1552)
+        // Therefore, constructor will not try to create the client
```
tomkerkhove: This should be moved to a method to perform the logic so that the client creation is done asynchronous, not synchronous

src/Arcus.Security.KeyVault/KeyVaultSecretProvider.cs
```diff
@@ -1,16 +1,69 @@
...
+        /// <returns>The value, stored in KeyVault</returns>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<string> Get(string name)
         {
-            throw new NotImplementedException();
+            Guard.NotNullOrEmpty(name, nameof(name));
+            try
+            {
+                SecretBundle secretBundle = await KeyVaultClient.GetSecretAsync(_keyVaultUri, name);
+                return secretBundle?.Value;
```
tomkerkhove: What if I want to know the version of the secret as well?

src/Arcus.Security.KeyVault/Factories/ServicePrincipalKeyVaultClientFactory.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.Threading.Tasks;
+using GuardNet;
+using Microsoft.Azure.KeyVault;
+using Microsoft.IdentityModel.Clients.ActiveDirectory;
+
+namespace Arcus.Security.KeyVault.Factories
+{
+    /// <summary>
+    /// <see cref="KeyVaultClientFactory"/> implementation using Azure Managed Service Identity
```
tomkerkhove: Copy paste

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            // Arrange
+            var clientId = Configuration.GetValue<string>("Arcus:Security:KeyVault:ClientId");
+            var clientKey = Configuration.GetValue<string>("Arcus:Security:KeyVault:ClientKey");
+            var keyVaultUri = Configuration.GetValue<string>("Arcus:Security:KeyVault:KeyVaultUri");
+            var keyName = Configuration.GetValue<string>("Arcus:Security:KeyVault:TestKeyName");
+
+            // Act
+            KeyVaultSecretProvider keyVaultSecretProvider = new KeyVaultSecretProvider(
+                new ServicePrincipalKeyVaultClientFactory(clientId, clientKey), keyVaultUri
```
tomkerkhove: Thinking out loud (and nit-picking) - Given the client factory can be different and also instantiated differently, maybe better to pass uri first?

src/Arcus.Security.KeyVault/KeyVaultSecretProvider.cs
```diff
@@ -1,16 +1,69 @@
...
+        /// <summary>
+        /// Creates an Azure KeyVault Secret provider, connected to a specific Azure Key Vault
+        /// </summary>
+        /// <param name="keyVaultClientFactory">A <see cref="KeyVaultClientFactory"/> implementation that will be used to generate the KeyVaultClient</param>
+        /// <param name="keyVaultUri">The Uri of the Azure KeyVault you want to connect to.  <example>https://{vaultname}nebulus-iot.vault.azure.net/</example></param>
+        public KeyVaultSecretProvider(KeyVaultClientFactory keyVaultClientFactory, string keyVaultUri)
+        {
+            Guard.NotNullOrEmpty(keyVaultUri, nameof(keyVaultUri));
+            Guard.NotNull(keyVaultClientFactory, nameof(keyVaultClientFactory));
+            _keyVaultUri = keyVaultUri;
```
tomkerkhove: Maybe interesting to guard on invalid uri by using `new Uri(keyVaultUri)`?

.gitignore
tomkerkhove: But I'm open for suggestions, how would you run the tests with this approach? Just rely on the environment variables?

The downside is that new people need to go through all tests to know what variables are required to run them locally.

tomkerkhove:
Feel free to also update the README/docs how to run the tests similar to Event Grid

.gitignore
SamVanhoutte: I agree it's not perfect and I thought I picked the same way of loading settings from the EventGrid integration tests.  Now, I just excluded the appsettings.json through gitignore, but people have to build their own way.  Will implement it in a new PR, coming up.

src/Arcus.Security.KeyVault/Factories/KeyVaultClientFactory.cs
SamVanhoutte: Because I think there might be common methods that can fit there.  I hesitated and went for the abstract class (which is also quite common in Factory patterns - at least in the good old times :))

src/Arcus.Security.Tests.Integration/Logging/XunitTestLogger.cs
SamVanhoutte: true - was copy/paste, as we might need this in multiple projects, it would be good to have these testing things centralized.  you mention Arcus.Core, but would it be valuable to have something like Arcus.Testing as well?  I'll leave it up to you to create the issue where you deem it fit 
Cheers

src/Arcus.Security.KeyVault/Factories/MsiKeyVaultClientFactory.cs
SamVanhoutte: fixed

.gitignore
SamVanhoutte: created #36

src/Arcus.Security.KeyVault/KeyVaultSecretProvider.cs
SamVanhoutte: fixed

src/Arcus.Security.KeyVault/KeyVaultSecretProvider.cs
SamVanhoutte: Can make a public property for this, but then it's still only available in the KeyVaultSpecifics.  Do you think we need a more generic approach for this? (what about other secretproviders?)

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
SamVanhoutte: I don't mind, but if you want to change it, better change it fast (or breaking)

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
SamVanhoutte: Sorry, created an issue :) #38

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
SamVanhoutte: agree

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
SamVanhoutte: What do you mean?  I have too much assertions in one test?

src/Arcus.Security.Tests.Integration/Logging/XunitTestLogger.cs
tomkerkhove: That would also be a good one, please create an issue on Arcus (general one) to keep track of this and we can move the conversation there - Will forget about it due to holidays :D

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: Not too many asserts but testing different scenarios. Anyway, jumped in and helped

src/Arcus.Security.Tests.Integration/Logging/XunitTestLogger.cs
tomkerkhove: Created https://github.com/arcus-azure/arcus/issues/19 for this

PR #41, opened 2018-12-19
CachedSecretProvider and related unit tests

SamVanhoutte:
Please review first if the approach is good.
After that, I'll create the docs

### DIFF
--- a/src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
+++ b/src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Arcus.Security.Core.Exceptions;
+using Arcus.Security.Core.Interfaces;
+using GuardNet;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace Arcus.Security.Core.Caching
+{
+    /// <inheritdoc cref="ISecretProvider"/>
+    /// <summary>
+    /// A Secret Provider that will cache secrets in memory, to improve performance
+    /// </summary>
+    public class CachedSecretProvider : ISecretProvider
+    {
+        private readonly ISecretProvider _secretProvider;
+        private readonly IMemoryCache _memoryCache;
+        private readonly TimeSpan _cacheDuration;
+
+        /// <summary>
+        /// Creating a new CachedSecretProvider with all required information
+        /// </summary>
+        /// <param name="secretProvider">The internal <see cref="ISecretProvider"/> used to retrieve the actual Secret Value, when not cached</param>
+        /// <param name="cacheDuration">The <see cref="TimeSpan"/> for which a retrieved value should be cached</param>
+        /// <param name="memoryCache">A <see cref="IMemoryCache"/> implementation that can cache data in memory.</param>
+        /// <exception cref="ArgumentNullException">The secretProvider and memoryCache parameters must not be null</exception>
+        public CachedSecretProvider(ISecretProvider secretProvider, TimeSpan cacheDuration, IMemoryCache memoryCache)
+        {
+            Guard.NotNull(secretProvider, nameof(secretProvider));
+            Guard.NotNull(memoryCache, nameof(memoryCache));
+
+            _secretProvider = secretProvider;
+            _memoryCache = memoryCache;
+            _cacheDuration = cacheDuration;
+        }
+
+        /// <summary>
+        /// Retrieves the secret value from the cache (when available) or from the internal ISecretProvider
+        /// </summary>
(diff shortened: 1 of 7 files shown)


tomkerkhove:
I'm not sure yet if we should go this route or pass an ICacheProvider to our secret providers who want to support it.

The latter approach would stimulate to use a cache while with this approach you need to know about it.

SamVanhoutte:
@tomkerkhove - agree with your statement
but wouldn't it be better to have an abstract class (`SecretProvider`) instead of `ISecretProvider ` , so that the actual caching logic only has to be done once (in the base class) - and others can just rely on it...

tomkerkhove:
> @tomkerkhove - agree with your statement
> but wouldn't it be better to have an abstract class (`SecretProvider`) instead of `ISecretProvider ` , so that the actual caching logic only has to be done once (in the base class) - and others can just rely on it...

I think a mixture would be perfect, ISecretProvider with SecretProvider (non-abstract) providing shared logic. Would that make sense?

Interfaces also help consumers using dependency injection.

However your approach is more along the "composition" approach which also makes sense, I'm a bit two-fold on this so not sure which one to pick.

src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+using GuardNet;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace Arcus.Security.Core.Caching
+{
+    /// <inheritdoc cref="ISecretProvider"/>
+    /// <summary>
+    /// A Secret Provider that will cache secrets in memory, to improve performance
+    /// </summary>
+    public class CachedSecretProvider : ISecretProvider
```
tomkerkhove: This should implement ICachedSecretProvider that enforces Get(name, skipCache:false) to force fresh gets in case of rolled secrets

SamVanhoutte: ok

tomkerkhove: Feel free to have a look at my key Vault rolling sample repo on Github

tomkerkhove:
Feel free to proceed as is
@SamVanhoutte and we can discuss after holiday

SamVanhoutte:
Introduced ICachedSecretProvider as requested and created an extension method to generate such a CachedVersion on existing ISecretProviders.  That way, it's not needed to repeat caching logic every time

src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name="secretProvider">The internal <see cref="ISecretProvider"/> used to retrieve the actual Secret Value, when not cached</param>
+        /// <param name="cacheDuration">The <see cref="TimeSpan"/> for which a retrieved value should be cached</param>
+        /// <param name="memoryCache">A <see cref="IMemoryCache"/> implementation that can cache data in memory.</param>
+        /// <exception cref="ArgumentNullException">The secretProvider and memoryCache parameters must not be null</exception>
+        public CachedSecretProvider(ISecretProvider secretProvider, TimeSpan cacheDuration, IMemoryCache memoryCache)
+        {
+            Guard.NotNull(secretProvider, nameof(secretProvider));
+
+            _secretProvider = secretProvider;
+            _memoryCache = memoryCache ?? new MemoryCache(new MemoryCacheOptions());
```
tomkerkhove: Maybe it's best to use overloads for this so that 1) our consumers know that it's possible to use it without their own cache 2) we can check for nulls if one is passed

SamVanhoutte: Done in new commit

src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        /// <summary>
+        /// Retrieves the secret value from the cache (when available) or from the internal ISecretProvider
+        /// </summary>
+        /// <param name="name">The name of the secret value to be retrieved</param>
+        /// <param name="skipCache">Indicates to skip the cache and force an update of the secret value</param>
+        /// <returns>The secret value</returns>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<string> Get(string name, bool skipCache)
```
tomkerkhove: Please take `skipCache` into account

SamVanhoutte:
Integration tests for Caching : all the logic is included in the unit tests already - where do you see the benefit for integration tests for the caching?  (as it is independent from Key Vault)

tomkerkhove:
I must've missed that - Sorry.

PR #45, opened 2019-01-04
 Modularize in decent package structure

tomkerkhove:
Modularize in decent package structure.

Currently we have the following structure:
arcus security packaging-page-2

This makes it easier for our consumers to only pull in the providers for what they want to achieve ie secrets. That said, we still have seperate packages for every single provider so if we add crypto, we can re-use the Key Vault shared core.

This change is done to have the following structure where we have multiple providers and different types such as secrets, crypto, etc.

arcus security packaging-proposal


SamVanhoutte:
The approach is okay for me and please go ahead with it.  I'm just not sure of the "Providers" name everywhere.  Maybe it's better, easier to leave the "providers" out of the package/namespace name?
But I'll leave it up to you

tomkerkhove:
Good point @SamVanhoutte !

Would this make more sense?
arcus security packaging-page-2 1


Note the removal of providers on package for consumer, but shared core still has it?

tomkerkhove:
We've agreed on this proposal and will make the changes asap.

Issue #8, opened 2018-10-08
Provide capability to retrieve and cache secret for Azure Key Vault


tomkerkhove:
Provide capability to cache the retrieved secret from #7.

**Describe the solution you'd like**
The client should be able to retrieve the secret from Azure Key Vault and use an in-memory cache to store it for a defined timespan.

This allows us to avoid hitting Azure Key Vault every time to avoid hitting the service limitations and avoid unrequired network calls.

tomkerkhove:
Can you provide docs for this please @SamVanhoutte?

PR #47, opened 2019-01-07
Introduce ICacheConfiguration

tomkerkhove:
Introduce ICacheConfiguration to make future changes easier.

Relates to #8


src/Arcus.Security.Core/Caching/Configuration/CacheConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        ///     Duration for which an entry should be cached
+        /// </summary>
+        public TimeSpan Duration { get; }
+
+        /// <summary>
+        ///     Constructor
+        /// </summary>
+        /// <param name="duration">Duration for which an entry should be cached</param>
```
SamVanhoutte: small suggestion: would be good to add <exception> in xmldocs

PR #51, opened 2019-01-11
FIX: make 'SecretNotFound' exception corectly serializable

stijnmoreels:
Exceptions requires a empty ctor to be correctly serializable.


stijnmoreels:
Yes, I feel the same way; 
I guess we need indeed 3 public ctors and add custom properties to the exception context. All the examples and documentation always has this. I even this about removing the `: base("secret not found...")` in the empty ctor since this isn't filled out by default in all the examples.

But I hear you. Apperently we can't use the _"Make invalid status unrepresentable"_ practice here.

src/Arcus.Security.Secrets.Core/Exceptions/SecretNotFoundException.cs
```diff
@@ -32,9 +40,32 @@ public SecretNotFoundException(string name, Exception innerException) : base($"T
...
+        /// <param name="context">The <see cref="T:StreamingContext"></see> that contains contextual information about the source or destination.</param>
+        protected SecretNotFoundException(SerializationInfo info, StreamingContext context) : base(info, context)
+        {
+            Name = info.GetString(nameof(Name));
+        }
+
         /// <summary>
         /// Name of the missing key
         /// </summary>
         public string Name { get; }
```
tomkerkhove: Let's add `Undefined` or something else that is explicit that it wasn't specified

stijnmoreels: Not sure what you mean? Updating the exception message?

tomkerkhove: No, assign a default value to `Name` with `Undefined` in case people call the default constructor

stijnmoreels: Aha, ok. Yes!

PR #52, opened 2019-01-14
FEAT: add client-throttling for TooManyRequests Key Vault response

stijnmoreels:
When the Key Vault service limits exceeds, we need to throttle the request to make sure we missue the Key Vault resources. This throttling happes when the Key Vault response has a HTTP status code 'Too Many Requests' (429) in the form of an exponential back-off.

#9


stijnmoreels:
Proposal (!). Not sure how or if I should test this because the method on the `KeyVaultClient.GetSecret(string, string, CancellationToken)` is an extension method + the factory we use doesn't return an abstraction but a concrete type.

See the several `// TODO`-comments I added to the file.

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -63,18 +77,43 @@ public async Task<string> Get(string secretName)
...
         private async Task<KeyVaultClient> GetClientAsync()
         {
-            if (_keyVaultClient == null)
+            await LockCreateKeyVaultClient.WaitAsync();
+
+            try
+            {
+                if (_keyVaultClient == null)
+                {
+                    // TODO: why is this factory not returning an interface (the 'Microsoft.Azure.KeyVault.IKeyVaultClient' interface)?
```
tomkerkhove: I don't think this matters much in this context to be honest

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -38,6 +43,10 @@ public KeyVaultSecretProvider(IKeyVaultAuthenticator authenticator, IKeyVaultCon
             VaultUri = $"{vaultConfiguration.VaultUri.Scheme}://{vaultConfiguration.VaultUri.Host}";
 
             _vaultConfiguration = vaultConfiguration;
+
+            /** TODO: if we only allow a single instance (KeyVaultClient) being created for this type,
```
tomkerkhove: Because those hand the instantiation which also defines how it authenticates

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -63,18 +77,43 @@ public async Task<string> Get(string secretName)
                     throw new SecretNotFoundException(secretName, keyVaultErrorException);
                 }
 
+                // TODO: are there other HTTP status codes that we can relate to secret exceptions?
```
tomkerkhove: Not that we are interested in

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -47,13 +56,18 @@ public KeyVaultSecretProvider(IKeyVaultAuthenticator authenticator, IKeyVaultCon
...
             try
             {
-                var keyVaultClient = await GetClientAsync();
-                var secretBundle = await keyVaultClient.GetSecretAsync(VaultUri, secretName);
+                KeyVaultClient keyVaultClient = await GetClientAsync();
+                SecretBundle secretBundle =
+                    await ThrottleTooManyRequests(
+                        () => keyVaultClient.GetSecretAsync(VaultUri, secretName));
+                
+                // TODO: should we allow to return 'null'?
```
tomkerkhove: If Key Vault says that the secret exist it's enough for us, it's not up to use to validate if the secret value is considered valid

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -63,18 +77,43 @@ public async Task<string> Get(string secretName)
...
+                return _keyVaultClient;
+            }
+            finally
             {
-                _keyVaultClient = await _authenticator.Authenticate();
+                LockCreateKeyVaultClient.Release();
             }
+        }
+
+        private static Task<SecretBundle> ThrottleTooManyRequests(Func<Task<SecretBundle>> getSecret)
```
tomkerkhove: Given this can be re-used for different operations we should consider renaming it to `secretOperation`

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: Ok, I'll remove this comment then.

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: Ok

stijnmoreels:
> Looks good, but please provide some unit testing with a stub or dummy (no mocking)

Yes, see my previous comments on how and if testing is required (re-testing only Polly).

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: See my previous comment(s)

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: See my previous comment(s)

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: Yes

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: Can you point out which one please?

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: Can you point out which one please?

tomkerkhove:
Testing is certainly possible and we will not re-test as it should be done from the consumer side to see if Polly is *used properly* over *testing Polly*

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: Thanks!

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: Merged in recent changes and this comment is no longer applicable I think

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: Don't really see the issue but I'm open to change. What would be your suggestion? To mock it probably?

tomkerkhove:
What would be your proposal then? Can you give some samples please?

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: Will be done after #55

Issue #55, opened 2019-03-04
Let the `IValueAuthenticator` respond with the `IKeyVaultClient` interface instead of the `KeyVaultClient` concrete type


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
As of today, the `IValueAuthenticator` returns a concrete `KeyVault` type which is later used to call the extension method `.GetSecretAsync(string, string)`. Because of this, there doesn't exist a testable hook to verify the newly added throttling functionality.

The pull request #52 had an issue when testing the `TooManyRequests` HTTP response from a key vault target.

**Describe the solution you'd like**
The solution could be changing the `IValueAuthenticator` to return an interface instead. This could be the existing `IKeyVault` interface.

Because of this change, we could without any problem create our own test double. The throttling functionality becomes testable.

**Describe alternatives you've considered**
We could use another method on the `KeyVault` client; but that is maybe not desirable.

PR #57, opened 2019-05-01
 FEAT: provide flexibility in 'KeyVaultSecretProvider' by using 'IKeyVaultCient'

stijnmoreels:
Because the 'IKeyVaultAuthenticator' used the concret 'KeyVaultClient',
we couldn't quite unit test the functionality.
The `KeyVaultSecretProvider` is now updated to use both the now obsolete
`IKeyVaultSecretAuthenticator` as the now new
`IKeyVaultSecretAuthentication` that uses the `IKeyVaultClient` as
return type.

This now allows reliable unit testing of the provider.

 #55

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
@@ -11,7 +11,7 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces
         /// <summary>
         ///     Authenticates with Azure Key Vault
         /// </summary>
-        /// <returns>A <see cref="KeyVaultClient" /> client to use for interaction with the vault</returns>
-        Task<IKeyVaultClient> Authenticate();
+        /// <returns>A <see cref="IKeyVaultClient" /> client to use for interaction with the vault</returns>
+        Task<IKeyVaultClient> AuthenticateClient();
     }
 }
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
@@ -1,17 +1,17 @@
 ﻿using System.Threading.Tasks;
 using Microsoft.Azure.KeyVault;
 
-namespace Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces
+namespace Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces 
 {
     /// <summary>
     ///     Authentication provider for Azure Key Vault
     /// </summary>
-    public interface IKeyVaultAuthenticator
+    public interface IKeyVaultAuthentication
     {
         /// <summary>
         ///     Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient" /> client to use for interaction with the vault</returns>
-        Task<KeyVaultClient> Authenticate();
+        Task<IKeyVaultClient> Authenticate();
     }
 }
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
@@ -12,6 +12,6 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces
         ///     Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="IKeyVaultClient" /> client to use for interaction with the vault</returns>
-        Task<IKeyVaultClient> AuthenticateClient();
+        Task<IKeyVaultClient> Authenticate();
     }
 }
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthenticator.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.KeyVault;
+
+namespace Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces
+{
+    /// <summary>
+    ///     Authentication provider for Azure Key Vault
+    /// </summary>
+    [Obsolete(
+        "Use the "
+        + nameof(IKeyVaultAuthentication)
+        + " instead to return an "
+        + nameof(IKeyVaultClient)
+        + " implementation instead of a concrete "
+        + nameof(KeyVaultClient))]
+    public interface IKeyVaultAuthenticator
+    {
+        /// <summary>
+        ///     Authenticates with Azure Key Vault
+        /// </summary>
+        /// <returns>A <see cref="KeyVaultClient" /> client to use for interaction with the vault</returns>
+        Task<KeyVaultClient> Authenticate();
+    }
+}
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
@@ -8,17 +8,17 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
     /// <summary>
     ///     Azure Key Vault authentication by using Azure Managed Service Identity
     /// </summary>
-    public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthenticator
+    public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthentication
     {
         /// <summary>
         /// Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient"/> client to use for interaction with the vault</returns>
-        public Task<KeyVaultClient> Authenticate()
+        public Task<IKeyVaultClient> Authenticate()
         {
             var tokenProvider = new AzureServiceTokenProvider();
             var authenticationCallback = new KeyVaultClient.AuthenticationCallback(tokenProvider.KeyVaultTokenCallback);
-            var keyVaultClient = new KeyVaultClient(authenticationCallback);
+            IKeyVaultClient keyVaultClient = new KeyVaultClient(authenticationCallback);
 
             return Task.FromResult(keyVaultClient);
         }
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
@@ -15,7 +15,7 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
         /// <summary>
         /// Authenticates with Azure Key Vault
         /// </summary>
-        /// <returns>A <see cref="KeyVaultClient"/> client to use for interaction with the vault</returns>
+        /// <returns>A <see cref="IKeyVaultClient"/> client to use for interaction with the vault</returns>
         public Task<IKeyVaultClient> Authenticate()
         {
             IKeyVaultClient client = AuthenticateClient();
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
@@ -16,23 +16,29 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
         /// Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient"/> client to use for interaction with the vault</returns>
-        public async Task<IKeyVaultClient> AuthenticateClient()
+        public Task<IKeyVaultClient> Authenticate()
         {
-            IKeyVaultClient client = await Authenticate();
-            return client;
+            IKeyVaultClient client = AuthenticateClient();
+            return Task.FromResult(client);
         }
 
         /// <summary>
         ///     Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient" /> client to use for interaction with the vault</returns>
-        public Task<KeyVaultClient> Authenticate()
+        Task<KeyVaultClient> IKeyVaultAuthenticator.Authenticate()
+        {
+            KeyVaultClient keyVaultClient = AuthenticateClient();
+            return Task.FromResult(keyVaultClient);
+        }
+
+        private static KeyVaultClient AuthenticateClient()
         {
             var tokenProvider = new AzureServiceTokenProvider();
             var authenticationCallback = new KeyVaultClient.AuthenticationCallback(tokenProvider.KeyVaultTokenCallback);
+            
             var keyVaultClient = new KeyVaultClient(authenticationCallback);
-
-            return Task.FromResult(keyVaultClient);
+            return keyVaultClient;
         }
     }
 }
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
@@ -8,17 +8,29 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
     /// <summary>
     ///     Azure Key Vault authentication by using Azure Managed Service Identity
     /// </summary>
-    public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthentication
+#pragma warning disable 618
+    public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthentication, IKeyVaultAuthenticator
+#pragma warning restore 618
     {
         /// <summary>
         /// Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient"/> client to use for interaction with the vault</returns>
-        public Task<IKeyVaultClient> Authenticate()
+        public async Task<IKeyVaultClient> AuthenticateClient()
+        {
+            IKeyVaultClient client = await Authenticate();
+            return client;
+        }
+
+        /// <summary>
+        ///     Authenticates with Azure Key Vault
+        /// </summary>
+        /// <returns>A <see cref="KeyVaultClient" /> client to use for interaction with the vault</returns>
+        public Task<KeyVaultClient> Authenticate()
         {
             var tokenProvider = new AzureServiceTokenProvider();
             var authenticationCallback = new KeyVaultClient.AuthenticationCallback(tokenProvider.KeyVaultTokenCallback);
-            IKeyVaultClient keyVaultClient = new KeyVaultClient(authenticationCallback);
+            var keyVaultClient = new KeyVaultClient(authenticationCallback);
 
             return Task.FromResult(keyVaultClient);
         }
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs
@@ -10,7 +10,9 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
     /// <summary>
     /// Representation of an <see cref="IKeyVaultAuthentication"/> that will generate a <see cref="IKeyVaultClient"/> implementation using a service principle.
     /// </summary>
-    public class ServicePrincipalAuthenticator : IKeyVaultAuthentication
+#pragma warning disable 618
+    public class ServicePrincipalAuthenticator : IKeyVaultAuthentication, IKeyVaultAuthenticator
+#pragma warning restore 618
     {
         private readonly string _clientId;
         private readonly string _clientKey;
@@ -33,9 +35,19 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
         /// Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient"/> client to use for interaction with the vault</returns>
-        public Task<IKeyVaultClient> Authenticate()
+        public Task<IKeyVaultClient> AuthenticateClient()
         {
             IKeyVaultClient keyVaultClient = new KeyVaultClient(GetToken);
+            return Task.FromResult(keyVaultClient);
+        }
+
+        /// <summary>
+        ///     Authenticates with Azure Key Vault
+        /// </summary>
+        /// <returns>A <see cref="KeyVaultClient" /> client to use for interaction with the vault</returns>
+        public Task<KeyVaultClient> Authenticate()
+        {
+            var keyVaultClient = new KeyVaultClient(GetToken);
             return Task.FromResult(keyVaultClient);
         }
 
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs
@@ -35,7 +35,7 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
         /// Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient"/> client to use for interaction with the vault</returns>
-        public Task<IKeyVaultClient> AuthenticateClient()
+        public Task<IKeyVaultClient> Authenticate()
         {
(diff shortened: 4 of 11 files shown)


src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
```diff
@@ -1,17 +1,17 @@
...
 using Microsoft.Azure.KeyVault;
 
-namespace Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces
+namespace Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces 
 {
     /// <summary>
     ///     Authentication provider for Azure Key Vault
     /// </summary>
-    public interface IKeyVaultAuthenticator
+    public interface IKeyVaultAuthentication
```
stijnmoreels: Apperently, this file was not in sync with the class; that's why it's renamed.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthenticator.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.KeyVault;
+
+namespace Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces
+{
+    /// <summary>
+    ///     Authentication provider for Azure Key Vault
+    /// </summary>
+    [Obsolete(
+        "Use the "
```
tomkerkhove: Why are we not just putting this in one line?

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs
```diff
@@ -7,13 +7,16 @@
 
 namespace Arcus.Security.Providers.AzureKeyVault.Authentication
 {
-    public class ServicePrincipalAuthenticator: IKeyVaultAuthenticator
+    /// <summary>
+    /// Representation of an <see cref="IKeyVaultAuthentication"/> that will generate a <see cref="IKeyVaultClient"/> implementation using a service principle.
+    /// </summary>
+    public class ServicePrincipalAuthenticator : IKeyVaultAuthentication
```
tomkerkhove: You'll need to implement both actually, otherwise, this is a breaking change

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
```diff
@@ -8,17 +8,17 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
     /// <summary>
     ///     Azure Key Vault authentication by using Azure Managed Service Identity
     /// </summary>
-    public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthenticator
+    public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthentication
```
tomkerkhove: You'll need to implement both actually, otherwise, this is a breaking change

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -35,18 +35,25 @@ public class KeyVaultSecretProvider : ISecretProvider
         /// </summary>
         /// <param name="authenticator">The requested authentication type for connecting to the Azure Key Vault instance</param>
         /// <param name="vaultConfiguration">Configuration related to the Azure Key Vault instance to use</param>
+        [Obsolete("Use other constructor with " + nameof(IKeyVaultAuthentication) + " instead")]
```
tomkerkhove: Maybe use $"" notation?

src/Arcus.Security.Tests.Unit/KeyVault/Doubles/AzureKeyVaultAuthenticatorDummy.cs
```diff
@@ -3,11 +3,11 @@
 using Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces;
 using Microsoft.Azure.KeyVault;
 
-namespace Arcus.Security.Tests.Unit.KeyVault.Dummies
+namespace Arcus.Security.Tests.Unit.KeyVault.Doubles
```
tomkerkhove: What's a Double? :)

src/Arcus.Security.Secrets.AzureKeyVault/CompatibleKeyVaultAuthentication.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            Guard.NotNull(authenticator, nameof(authenticator));
+            
+            _authenticator = authenticator;
+        }
+
+        /// <summary>
+        ///     Authenticates with Azure Key Vault
+        /// </summary>
+        /// <returns>A <see cref="KeyVaultClient" /> client to use for interaction with the vault</returns>
```
tomkerkhove: Returns interface, not concrete class

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -25,38 +31,71 @@ public void KeyVaultSecretProvider_CreateWithHttpScheme_ShouldFailWithUriFormatE
...
 
         [Fact]
         public void KeyVaultSecretProvider_CreateWithoutUri_ShouldFailWithArgumentException()
         {
-            // Arrange
-            string uri = null;
-
             // Act & Assert
-            Assert.ThrowsAny<ArgumentException>(() => new KeyVaultSecretProvider(new AzureKeyVaultAuthenticatorDummy(), new KeyVaultConfiguration(uri)));
+            Assert.ThrowsAny<ArgumentException>(
```
tomkerkhove: What's the added value of this?

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -25,38 +31,71 @@ public void KeyVaultSecretProvider_CreateWithHttpScheme_ShouldFailWithUriFormatE
...
+                    new AzureKeyVaultAuthenticatorDummy(), 
+                    new KeyVaultConfiguration(rawVaultUri: null)));
         }
 
         [Fact]
         public void KeyVaultSecretProvider_CreateWithoutClientFactory_ShouldFailWithArgumentException()
         {
             // Arrange
-            string uri = $"https://{Guid.NewGuid():N}.vault.azure.net/";
+            string uri = GenerateVaultUri();
```
tomkerkhove: Let's rename to `GenerateBogusVaultUri` maybe?

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -25,38 +31,71 @@ public void KeyVaultSecretProvider_CreateWithHttpScheme_ShouldFailWithUriFormatE
...
+            string expected = $"secret-value-{Guid.NewGuid()}";
+
+            var keyVaultClient = new SimulatedKeyVaultClient(
+                request => throw new KeyVaultErrorException("Sabotage secret retrieval with TooManyRequests")
+                {
+                    Response = new HttpResponseMessageWrapper(
+                        new HttpResponseMessage(HttpStatusCode.TooManyRequests), 
+                        "some HTTP response content to ignore")
+                },
+                request => new SecretBundle(value: expected));
```
tomkerkhove: Since we don't need request here, can't we just take in a bundle?

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -25,38 +31,71 @@ public void KeyVaultSecretProvider_CreateWithHttpScheme_ShouldFailWithUriFormatE
...
+
+        [Fact]
+        public async Task KeyVaultSecretProvider_GetsSecretValue_AfterRetriedTooManyRequestException()
+        {
+            // Arrange
+            string secretName = $"secret-name-{Guid.NewGuid()}";
+            string expected = $"secret-value-{Guid.NewGuid()}";
+
+            var keyVaultClient = new SimulatedKeyVaultClient(
+                request => throw new KeyVaultErrorException("Sabotage secret retrieval with TooManyRequests")
```
tomkerkhove: It must be my ignorance, but this is very strange syntax

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: Can't; should be const `string`.

src/Arcus.Security.Tests.Unit/KeyVault/Doubles/AzureKeyVaultAuthenticatorDummy.cs
stijnmoreels: A _Test Double_ is a concept that's an umbrella term for all types created to manipulate test input: http://xunitpatterns.com/Test%20Double.html

src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthenticator.cs
stijnmoreels: It was otherwise very long line to read; and although I thought it would make it less readable to place it separatly; I liked it more this way.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
stijnmoreels: Ah, yes; hmm; then I'll need to rename `Authenticate`.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthenticator.cs
stijnmoreels: It was otherwise very long line to read; and although I thought it would make it less readable to place it separatly; I liked it more this way.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs
stijnmoreels: Ah, yes; hmm; then I have to rename `Authenticate`.

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: Yes, but I thought if we would like to black-box test indirect outputs from that we send via the key secret provider; we would have the same values as in the simulated key vault client.
Seems to me that this makes it easier to understand what the simulated client does.

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: Strange in what way?

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: Is it actualy a generated bogus? We're not explicitly testing URI parsing and/or input validation; so I guess it would make it less obvious?

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: Formatting you mean? it helps when you place each separate argument on a single line so you don't confuse which argument is sent to which instance.

src/Arcus.Security.Secrets.AzureKeyVault/CompatibleKeyVaultAuthentication.cs
stijnmoreels: Ah yes, thanks.

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: Sure, but do we actually need it now? If not, I would keep it simple - The YAGNI way

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: 1. A lot of stuff which is done in constructor call
2. THrow exception after which we run a lambda
3. Request not being used

image

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
stijnmoreels: Renamed it to `AuthenticateClient`

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: It's actually a series of simulations (functions) that we should run in sequence for each request.
It's a very common way to use it this way but with the right suggestions I can make it more OOP?

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: Ah, yes. OK

src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthenticator.cs
tomkerkhove: I would move it all in one line as it creates a bit of noise but that's ok I guess.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
```diff
@@ -8,12 +8,24 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
...
     /// </summary>
-    public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthenticator
+#pragma warning disable 618
+    public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthentication, IKeyVaultAuthenticator
+#pragma warning restore 618
     {
         /// <summary>
         /// Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient"/> client to use for interaction with the vault</returns>
```
tomkerkhove:
```suggestion
        /// <returns>A <see cref="IKeyVaultClient"/> client to use for interaction with the vault</returns>
```

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
```diff
@@ -8,12 +8,24 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
...
+        public async Task<IKeyVaultClient> AuthenticateClient()
+        {
+            IKeyVaultClient client = await Authenticate();
+            return client;
+        }
+
+        /// <summary>
+        ///     Authenticates with Azure Key Vault
```
tomkerkhove: Let's align formatting with current docs or update the current ones with this approach

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs
```diff
@@ -30,6 +35,16 @@ public ServicePrincipalAuthenticator(string clientId, string clientKey)
         /// Authenticates with Azure Key Vault
         /// </summary>
         /// <returns>A <see cref="KeyVaultClient"/> client to use for interaction with the vault</returns>
```
tomkerkhove:
```suggestion
        /// <returns>A <see cref="IKeyVaultClient"/> client to use for interaction with the vault</returns>
```

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
tomkerkhove: This goes for all others as well

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -35,19 +35,26 @@ public class KeyVaultSecretProvider : ISecretProvider
...
+#pragma warning disable 618
         public KeyVaultSecretProvider(IKeyVaultAuthenticator authenticator, IKeyVaultConfiguration vaultConfiguration)
+            : this(new CompatibleKeyVaultAuthentication(authenticator), vaultConfiguration) { }
```
tomkerkhove: We should be consistent in our formatting
```suggestion
            : this(new CompatibleKeyVaultAuthentication(authenticator), vaultConfiguration)
            {
            }
```

src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
```diff
@@ -1,17 +1,17 @@
...
     {
         /// <summary>
         ///     Authenticates with Azure Key Vault
         /// </summary>
-        /// <returns>A <see cref="KeyVaultClient" /> client to use for interaction with the vault</returns>
-        Task<KeyVaultClient> Authenticate();
+        /// <returns>A <see cref="IKeyVaultClient" /> client to use for interaction with the vault</returns>
+        Task<IKeyVaultClient> AuthenticateClient();
```
tomkerkhove: Why don't we just call it Authenticate like it was before? :)
```suggestion
        Task<IKeyVaultClient> Authenticate();
```

tomkerkhove: You should still be able to use IKeyVaultAuthentication.Authenticate if it's for implementing this interface

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: Can you elaborate on that please? I'm not really sure I get the issue as intellisense would help when adding new parameters but I'm probably not getting the point

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: I'm using bogus as it does not exist, not perse a reference to the libarary. Feels like you'd expect a valid vault URI here imo, but was just an idea

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
tomkerkhove: Noticed this, don't really see the need why we need to rename it to be honest.

Idea is that it authenticates, doesn't need to imply that it gives you a client imo.

I think we should keep `Authenticate`

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs
tomkerkhove: Not really

src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
stijnmoreels: Because it would conflict with the obsolete interface

tomkerkhove: I don't see how. You can implement them both with InterfaceX.Authenticate & InterfaceY.Authenticate

stijnmoreels: Yes, I know that you can get this working with explicit interface implementation, but that also means that you have to cast the instance to the correct interface if you want to call a specific `Authenticate`.
Which means that you could get scenario's where you aren't exactly fully compatible and would get errors while compiling.

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: No, it's not a point of intellesense or programmability; it's about readability. When all arguments are placed on the same line you can't tell in one instance which argument is sent to which instance because you first have to look where each paranthesis start and ends.
Placing them on a seperate line helps with seeing the structure in which an instance is build up.

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: :)

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: I think we don't need that but ok.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
tomkerkhove: Given we only call it internally based on the interfaces I think we will not have issues, what am I missing? :)

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: You're throwing an exception after which you run a function, not really sure why we are using this kind of syntax, this feels strange.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/Interfaces/IKeyVaultAuthentication.cs
stijnmoreels: Ah, OK. If you're sure that we only use this internally.

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: 1. Not entirely correct. The ctor gets a list of simulations which it runs in sequence. The functions are therfore "lazy" executed.
2. This syntax allows the test reader to see the order in which the key vault client will act on incoming requests.
3. It may feel strange because this way of working is using functions as first class citizens just like any other value; which is the core principle of functional programming. But from all syntaxes I've seen to test this; this feels to me as the most readable and maintainable solution.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
stijnmoreels: Yes, changed.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
tomkerkhove: It's ok for now

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: But we are not using functional programming on this project so it does not make sense to shift to that flavor imo.

Anyway, business-wise it's good enough and don't want to keep this PR open any longer than needed so I'm resolving this.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
tomkerkhove: Was going to merge this one but I can't, feel free to commit this suggestion if you agree @stijnmoreels

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: :D That's funny you say that. LINQ, async/await, delegates/lamdba, ... are all functional concepts. So you should't be using those constructs either. If you're using a recent version of C#, you're almost always using some functional programming ideas.

Issue #37, opened 2018-12-18
Expose Version of Key Vault Secret in KeyVaultSecretProvider


SamVanhoutte:
In the KeyVaultSecretProvider, we return the value of a secret, when found
At the same time, it can be needed to know the version of the secret that was retrieved.

This issue is to track it

stijnmoreels:
This will change the `ISecretProvider`, @tomkerkhove this is probably for later version?

tomkerkhove:
You can already add a new version next to it ;)

stijnmoreels:
I propose creating a new model that has value+version so we don't depend on microsoft models?
@tomkerkhove what is the default approach on Arcus for obsolete functionality: versioned interface + obsolete attribute?

tomkerkhove:
You don't need a versioned interface imo, you can just add the new method next to the current one.

I'm not sure if we should add our own model given people want to have more information about the secret itself as well, but we can start with our own model until there is effective demand for this.

stijnmoreels:
> You don't need a versioned interface imo, you can just add the new method next to the current one.

So you mean that we should allow a breaking change on the next release?

> I'm not sure if we should add our own model given people want to have more information about the secret itself as well, but we can start with our own model until there is effective demand for this.

Ok

tomkerkhove:
I'm saying that we can add the new endpoint to the current interface and mark the old one as obsolete. Before we ship 1.0.0 we'll get rid of that one then.

Feel free to work on this if you want @stijnmoreels and as suggested by you a small custom model is ok.

tomkerkhove:
Done via #58

PR #58, opened 2019-05-15
FEAT: expose secret version in 'ISecretProvider'

stijnmoreels:
Because further clients will need the version of the secret, this should
be exposed together with the secret value itself.

 #37

### DIFF
--- a/src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
+++ b/src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
@@ -67,7 +67,21 @@ namespace Arcus.Security.Secrets.AzureKeyVault
         /// <returns>The value, stored in Key Vault</returns>
         /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
         /// <exception cref="KeyVaultErrorException">The call for a secret resulted in an invalid response</exception>
-        public async Task<string> Get(string secretName)
+        [Obsolete("Use the " + nameof(GetRawSecret) + " method instead")]
+        public Task<string> Get(string secretName)
+        {
+            return GetRawSecret(secretName);
+        }
+
+        /// <summary>
+        /// Retrieves the secret value, based on the given name
+        /// </summary>
+        /// <param name="secretName">The name of the secret key</param>
+        /// <returns>Returns the secret key.</returns>
+        /// <exception cref="ArgumentException">The name must not be empty</exception>
+        /// <exception cref="ArgumentNullException">The name must not be null</exception>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<string> GetRawSecret(string secretName)
         {
             Secret secret = await GetSecret(secretName);
             return secret?.Value;
--- a/src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
+++ b/src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
@@ -14,44 +14,53 @@ using Polly;
 namespace Arcus.Security.Secrets.AzureKeyVault
 {
     /// <summary>
-    /// Secret key provider that connects to Azure Key Vault
+    ///     Secret key provider that connects to Azure Key Vault
     /// </summary>
     public class KeyVaultSecretProvider : ISecretProvider
     {
-        private readonly IKeyVaultAuthenticator _authenticator;
+        private readonly IKeyVaultAuthentication _authentication;
         private readonly IKeyVaultConfiguration _vaultConfiguration;
 
-        private KeyVaultClient _keyVaultClient;
+        private IKeyVaultClient _keyVaultClient;
 
         private static readonly SemaphoreSlim LockCreateKeyVaultClient = new SemaphoreSlim(initialCount: 1, maxCount: 1);
 
         /// <summary>
-        /// Uri of the vault
+        ///     Uri of the vault
         /// </summary>
         public string VaultUri { get; }
 
         /// <summary>
-        /// Creates an Azure Key Vault Secret provider, connected to a specific Azure Key Vault
+        ///     Creates an Azure Key Vault Secret provider, connected to a specific Azure Key Vault
         /// </summary>
         /// <param name="authenticator">The requested authentication type for connecting to the Azure Key Vault instance</param>
         /// <param name="vaultConfiguration">Configuration related to the Azure Key Vault instance to use</param>
+        [Obsolete("Use other constructor with " + nameof(IKeyVaultAuthentication) + " instead")]
+#pragma warning disable 618
         public KeyVaultSecretProvider(IKeyVaultAuthenticator authenticator, IKeyVaultConfiguration vaultConfiguration)
+            : this(new CompatibleKeyVaultAuthentication(authenticator), vaultConfiguration)
+            {
+            }
+#pragma warning restore 618
+
+        /// <summary>
+        ///     Creates an Azure Key Vault Secret provider, connected to a specific Azure Key Vault
+        /// </summary>
+        /// <param name="authentication">The requested authentication type for connecting to the Azure Key Vault instance</param>
+        /// <param name="vaultConfiguration">Configuration related to the Azure Key Vault instance to use</param>
+        public KeyVaultSecretProvider(IKeyVaultAuthentication authentication, IKeyVaultConfiguration vaultConfiguration)
         {
             Guard.NotNull(vaultConfiguration, nameof(vaultConfiguration));
-            Guard.NotNull(authenticator, nameof(authenticator));
+            Guard.NotNull(authentication, nameof(authentication));
 
             VaultUri = $"{vaultConfiguration.VaultUri.Scheme}://{vaultConfiguration.VaultUri.Host}";
 
             _vaultConfiguration = vaultConfiguration;
-
-            /** TODO: if we only allow a single instance (KeyVaultClient) being created for this type,
-             *       why shouldn't we then take a dependency on the client instead on the factory?
-             *       The factory can still create the instance for us, but this provider may not be the right responsible to call this creation. */
-            _authenticator = authenticator;
+            _authentication = authentication;
         }
 
         /// <summary>
-        /// Gets the secret from Key Vault, using the right secret name
+        ///     Gets the secret from Key Vault, using the right secret name
         /// </summary>
         /// <param name="secretName">The secret name</param>
         /// <returns>The value, stored in Key Vault</returns>
@@ -62,7 +71,7 @@ namespace Arcus.Security.Secrets.AzureKeyVault
             Guard.NotNullOrEmpty(secretName, nameof(secretName));
             try
             {
-                KeyVaultClient keyVaultClient = await GetClientAsync();
+                IKeyVaultClient keyVaultClient = await GetClientAsync();
                 SecretBundle secretBundle =
                     await ThrottleTooManyRequests(
                         () => keyVaultClient.GetSecretAsync(VaultUri, secretName));
@@ -80,7 +89,7 @@ namespace Arcus.Security.Secrets.AzureKeyVault
             }
         }
 
-        private async Task<KeyVaultClient> GetClientAsync()
+        private async Task<IKeyVaultClient> GetClientAsync()
         {
             await LockCreateKeyVaultClient.WaitAsync();
 
@@ -88,8 +97,7 @@ namespace Arcus.Security.Secrets.AzureKeyVault
             {
                 if (_keyVaultClient == null)
                 {
-                    // TODO: why is this factory not returning an interface (the 'Microsoft.Azure.KeyVault.IKeyVaultClient' interface)?
-                    _keyVaultClient = await _authenticator.Authenticate();
+                    _keyVaultClient = await _authentication.Authenticate();
                 }
 
                 return _keyVaultClient;
@@ -102,7 +110,7 @@ namespace Arcus.Security.Secrets.AzureKeyVault
 
         private static Task<SecretBundle> ThrottleTooManyRequests(Func<Task<SecretBundle>> secretOperation)
         {
-            /** Client-side throttling using exponential back-off when Key Vault service limit exceeds:
+            /* Client-side throttling using exponential back-off when Key Vault service limit exceeds:
              * 1. Wait 1 second, retry request
              * 2. If still throttled wait 2 seconds, retry request
              * 3. If still throttled wait 4 seconds, retry request
--- a/src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
+++ b/src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
@@ -6,6 +6,7 @@ using Arcus.Security.Providers.AzureKeyVault.Authentication.Interfaces;
 using Arcus.Security.Providers.AzureKeyVault.Configuration.Interfaces;
 using Arcus.Security.Secrets.Core.Exceptions;
 using Arcus.Security.Secrets.Core.Interfaces;
+using Arcus.Security.Secrets.Core.Models;
 using GuardNet;
 using Microsoft.Azure.KeyVault;
 using Microsoft.Azure.KeyVault.Models;
@@ -68,6 +69,20 @@ namespace Arcus.Security.Secrets.AzureKeyVault
         /// <exception cref="KeyVaultErrorException">The call for a secret resulted in an invalid response</exception>
         public async Task<string> Get(string secretName)
         {
+            Secret secret = await GetSecret(secretName);
+            return secret?.Value;
+        }
+
+        /// <summary>
+        /// Retrieves the secret value, based on the given name
+        /// </summary>
+        /// <param name="secretName">The name of the secret key</param>
+        /// <returns>Returns a <see cref="Secret"/> that contains the secret key</returns>
+        /// <exception cref="ArgumentException">The name must not be empty</exception>
+        /// <exception cref="ArgumentNullException">The name must not be null</exception>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<Secret> GetSecret(string secretName)
+        {
             Guard.NotNullOrEmpty(secretName, nameof(secretName));
             try
             {
@@ -75,8 +90,13 @@ namespace Arcus.Security.Secrets.AzureKeyVault
                 SecretBundle secretBundle =
                     await ThrottleTooManyRequests(
                         () => keyVaultClient.GetSecretAsync(VaultUri, secretName));
-                
-                return secretBundle?.Value;
+
+                if (secretBundle == null)
+                {
+                    return null;
+                }
+
+                return new Secret(secretBundle.Value, secretBundle.SecretIdentifier?.Version);
             }
             catch (KeyVaultErrorException keyVaultErrorException)
             {
--- a/src/Arcus.Security.Secrets.Core/Interfaces/ISecretProvider.cs
+++ b/src/Arcus.Security.Secrets.Core/Interfaces/ISecretProvider.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Threading.Tasks;
 using Arcus.Security.Secrets.Core.Exceptions;
+using Arcus.Security.Secrets.Core.Models;
 
 namespace Arcus.Security.Secrets.Core.Interfaces
 {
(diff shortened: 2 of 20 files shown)


stijnmoreels:
I'm still no fan of categorizing the classes by type and not by functionality but I guess this was something in the code style guidlines?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -95,7 +96,28 @@ public async Task CachedSecretProvider_TwoCallsWithinCacheInterval_ShouldGetSame
...
+            string keyName = "MyValue";
+
+            // Act 
+            ICachedSecretProvider cachedSecretProvider = new CachedSecretProvider(testSecretProvider, cacheConfiguration, memCache);
+            var firstSecret = await cachedSecretProvider.GetSecret(keyName);
+            testSecretProvider.SecretValue = Guid.NewGuid().ToString("N"); // Change actual value on the internal secret provider !
+            var secondSecret = await cachedSecretProvider.GetSecret(keyName);
+
+            // Assert
+            Assert.Equal(firstSecret, secondSecret);
```
tomkerkhove: Maybe also assert the value & secret?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -119,7 +141,34 @@ public async Task CachedSecretProvider_TwoCallsOutsideCacheInterval_ShouldGetDif
...
+            // Act 
+            ICachedSecretProvider cachedSecretProvider = new CachedSecretProvider(testSecretProvider, cacheConfiguration, memCache);
+            Secret firstSecret = await cachedSecretProvider.GetSecret(keyName);
+            await Task.Delay(TimeSpan.FromMilliseconds(150));
+            string newSecretValue = Guid.NewGuid().ToString("N");
+            testSecretProvider.SecretValue = newSecretValue; // Change actual value on the internal secret provider !
+            Secret secondSecret = await cachedSecretProvider.GetSecret(keyName);
+
+            // Assert
+            Assert.True(firstSecret != null, "firstSecret != null");
```
tomkerkhove: Use Assert.NotNull or how is it called?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -119,7 +141,34 @@ public async Task CachedSecretProvider_TwoCallsOutsideCacheInterval_ShouldGetDif
...
+            Assert.True(secondSecret != null, "secondSecret != null");
+
+            Assert.Equal(secretKeyValue, firstSecret.Value);
```
tomkerkhove: Maybe also assert the value & secret?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -141,5 +190,32 @@ public async Task CachedSecretProvider_SkipCached_ShouldGetDifferentValue()
...
+            // Act 
+            ICachedSecretProvider cachedSecretProvider = new CachedSecretProvider(testSecretProvider, cacheConfiguration, memCache);
+            Secret firstSecret = await cachedSecretProvider.GetSecret(keyName);
+            await Task.Delay(TimeSpan.FromMilliseconds(150));
+            string newSecretValue = Guid.NewGuid().ToString("N");
+            testSecretProvider.SecretValue = newSecretValue; // Change actual value on the internal secret provider !
+            Secret secondSecret = await cachedSecretProvider.GetSecret(keyName, true);
+
+            // Assert
+            Assert.True(firstSecret != null, "firstSecret != null");
```
tomkerkhove: Same here

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -141,5 +190,32 @@ public async Task CachedSecretProvider_SkipCached_ShouldGetDifferentValue()
...
+            Assert.True(secondSecret != null, "secondSecret != null");
+
+            Assert.Equal(secretKeyValue, firstSecret.Value);
```
tomkerkhove: Maybe also assert the value & secret?

src/Arcus.Security.Tests.Unit/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -93,6 +95,35 @@ public async Task KeyVaultSecretProvider_GetsSecretValue_AfterRetriedTooManyRequ
...
+            var provider = new KeyVaultSecretProvider(
+                new StubKeyVaultAuthenticator(keyVaultClient), 
+                new KeyVaultConfiguration(GenerateVaultUri()));
+
+            // Act
+            Secret actual = await provider.GetSecret(secretName);
+
+            // Assert
+            Assert.NotNull(actual);
+            Assert.Equal(expected, actual.Value);
```
tomkerkhove: How about version? Maybe interesting to add this as well?

src/Arcus.Security.Secrets.Core/Interfaces/ISecretProvider.cs
```diff
@@ -17,6 +18,17 @@ public interface ISecretProvider
         /// <exception cref="ArgumentException">The name must not be empty</exception>
         /// <exception cref="ArgumentNullException">The name must not be null</exception>
         /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        [Obsolete("Use the " + nameof(GetSecret) + " method instead.")]
```
tomkerkhove: Actually, I'm wondering - Do we actually need to deprecate this one?

The only reason why I would remove it maybe is that we are introducing `SecretProvider.GetSecret` but that's fair.

What do you think?

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -67,6 +68,20 @@ public KeyVaultSecretProvider(IKeyVaultAuthentication authentication, IKeyVaultC
...
+
+        /// <summary>
+        /// Retrieves the secret value, based on the given name
+        /// </summary>
+        /// <param name="secretName">The name of the secret key</param>
+        /// <returns>Returns a <see cref="Secret"/> that contains the secret key</returns>
+        /// <exception cref="ArgumentException">The name must not be empty</exception>
+        /// <exception cref="ArgumentNullException">The name must not be null</exception>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<Secret> GetSecret(string secretName)
```
tomkerkhove: Wondering if we should return Secret or a monad.

src/Arcus.Security.Secrets.Core/Models/Secret.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Represents the secret returned from the <see cref="ISecretProvider"/> implementation.
+    /// </summary>
+    public class Secret
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Secret"/> class.
+        /// </summary>
+        /// <param name="value">The secret value.</param>
+        /// <param name="version">The version of the secret.</param>
+        public Secret(string value, string version)
```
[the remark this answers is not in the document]
stijnmoreels: Yes, not sure if this version is always set. Can't set it (without reflection) in tests also.

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: We're propably talking about different stuff here when we say "monad" (because `Task` is a monad in category theory/functional programming).

src/Arcus.Security.Secrets.Core/Interfaces/ISecretProvider.cs
stijnmoreels: For easier access maybe to only get the secret value.
Maybe rename it from `.Get()` to `.GetSecretValue()` then?

src/Arcus.Security.Secrets.Core/Models/Secret.cs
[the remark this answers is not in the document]
stijnmoreels: Ok, inspected the `SecretBundle` and `SecretIdentifier` and the way the `Version` gets set.
Let me change that for you.

src/Arcus.Security.Secrets.Core/Interfaces/ISecretProvider.cs
tomkerkhove: That's actually a breaking change, let's keep this one deprecated but add these new endpoints in this PR:
- `GetSecret`
- `GetRawSecret` which is new implementation of Get

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: Probably. I'm talking about the Maybe option
https://mikhail.io/2016/01/monads-explained-in-csharp/

stijnmoreels: Ah, ok; than we're talking about the same thing.

src/Arcus.Security.Secrets.Core/Interfaces/ISecretProvider.cs
stijnmoreels: Yes, I know; but the change you requested (adding a `.GetSecret` method to the interface) is also a breaking change to any clients?

tomkerkhove: You mean because it's a new method on an interface or?

src/Arcus.Security.Secrets.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: The task is a monad, but not the object that it returns. Anyway, let's skip it for now then.

src/Arcus.Security.Secrets.Core/Interfaces/ISecretProvider.cs
tomkerkhove: If so, I think it's ok to consider this a change that is ok to make imo

stijnmoreels: So we could add an extra method to the interface but could not rename the existing one?

stijnmoreels: So we could add an extra method to the interface but could not rename the existing one?

stijnmoreels: So we could add an extra method to the interface but could not rename the existing one?

tomkerkhove: Yes as adding will only break implementing classes. Rename will break implementing and consuming

stijnmoreels: Ah yes OK I see. Thanks for clearing that up.

stijnmoreels: > * `GetSecret`
> * `GetRawSecret` which is new implementation of Get

Ok, I go with your previous example!

tomkerkhove:
Just a few more comments and we're good to go!

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -21,44 +22,85 @@ public KeyVaultSecretProviderTests(ITestOutputHelper testOutput) : base(testOutp
...
+            
+            var keyVaultSecretProvider = new KeyVaultSecretProvider(
+                authentication: new ServicePrincipalAuthenticator(clientId, clientKey), 
+                vaultConfiguration: new KeyVaultConfiguration(keyVaultUri));
+
+            // Act
+            Secret secret = await keyVaultSecretProvider.GetSecret(keyName);
+
+            // Assert
+            Assert.NotNull(secret);
```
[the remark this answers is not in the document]
stijnmoreels: But since this is handled in the `Secret` class itself, it's rather strange to test on this?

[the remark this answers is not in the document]
tomkerkhove: I as a consumer am not aware of this and should not care about the implementation but more about the test outcome for this integration test in my opinion.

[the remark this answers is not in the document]
stijnmoreels: Ok, I'll do that; but doesn't make much sense for me.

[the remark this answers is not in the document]
tomkerkhove: Why does this not make sense to you?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -56,7 +55,7 @@ public void CachedSecretProvider_CreateWithoutCache_ShouldSucceed()
             var cacheConfiguration = new CacheConfiguration(TimeSpan.MaxValue);
 
             // Act & Assert
-            new CachedSecretProvider(testSecretProvider, cacheConfiguration);
+            var _ = new CachedSecretProvider(testSecretProvider, cacheConfiguration);
```
tomkerkhove: Just curious, what is the point of this?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -44,7 +43,7 @@ public void CachedSecretProvider_CreateWithOnlySecretProvider_ShouldSucceed()
             var testSecretProvider = new TestSecretProviderStub(secretKeyValue);
 
             // Act & Assert
-            var cachedSecretProvider = new CachedSecretProvider(testSecretProvider);
+            var _ = new CachedSecretProvider(testSecretProvider);
```
tomkerkhove: Just curious, what is the point of this?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: That was done.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: I suggest not to; because then you wouldn't know which once failled. Better for _Defect Localization_.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: That was done.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: That was done.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: That was done.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: Some practice I picked up; that you should always provide a value for the ctor to make it more readable.
I wonder though what the point was that was tested here...

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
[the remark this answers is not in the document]
stijnmoreels: "_Make invalid sates unrepresentable_" is always my goal. We could test on the 'Secret'  itself though (separate tests) but not the usage of the 'Secret' instance.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
tomkerkhove: I don't think we need to capture the response here in a variable.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
tomkerkhove: I don't think we need to capture the result here in a variable.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
tomkerkhove: It doesn't matter actually, I just don't like `_` names but let's leave it as is.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -1,10 +1,9 @@
 ﻿using System;
```
tomkerkhove: Frankly, this is how you can go from readable tests to unreadable tests imo.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -73,73 +72,171 @@ public void CachedSecretProvider_CreateWithCorrectArguments_ShouldSucceed()
...
-            var secondValue = await cachedSecretProvider.Get(keyName);
+            (string actualFirstSecret, string actualSecondSecret) = await GetTwoSecretsFromCachedProvider(
+                (expectedFirstSecret, expectedSecondSecret),
+                (provider, keyName) => provider.Get(keyName),
+                cacheSetup);
 
             // Assert
-            Assert.Equal(firstValue, secondValue);
-            Assert.Equal(1, testSecretProvider.CallsMadeSinceCreation);
+            Assert.True(
```
tomkerkhove: If you need to add checks like `cacheSetup == GetCachedSecret.WithinCacheInterval` is inline data the best approach? I'd argue that this means that it should have a dedicated test case.

Certainly if the second assert starts with `cacheSetup == GetCachedSecret.OutsideCacheInterval ` and the 3rd scenario will thus not have a assertion.

stijnmoreels: :) this approach is used quite a lot in property-based testing because of it generic usability. By abstracting the setup in readable values (enums) we quickly spot that this test file contains three different "properties" that will test three different system-endpoints.

The assertion you describe, is actually the idea behind "test-properties" that you define a relationship between test-input and test-output.

`cacheSetup == WithinCacheInterval == (actual1st == actual2st && expect1st == actual1st)`
becomes "_when the secret is retrieved within the cache interval, the same secret will be returned_".
Which is what we in fact want to test.

tomkerkhove: I certainly get why you are taking this approach and appreciate the suggestion, but I'm afraid this will decrease the accessibility for new folks on the project.

Personally, I think it's better to stick with the current approach where we create the objects in every test, do the logic and have simple asserts.

If the setup is different, and thus requires different asserting, then that means we need a separate test case.

Would it be possible to make the changes accordingly, please?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
tomkerkhove: Hm I don't think I'm following, can you elaborate on `then you wouldn't know which once failled` please?

Are you referring to `firstSecret` vs `secondSecret`? If so, the description of the assert should be able to help here.

stijnmoreels: Yes, but only `Assert.True/False` allows a custom `userMessage` to be passed. :(

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: Ok, I look into this.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
tomkerkhove: Oh I love NUnit more than xUnit, this sucks. Ok then :(

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
tomkerkhove: Thanks!

stijnmoreels: Ok, this is done I think?

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -72,171 +72,265 @@ public void CachedSecretProvider_CreateWithCorrectArguments_ShouldSucceed()
...
+
+            const string keyName = "MyValue";
+            ICachedSecretProvider cachedSecretProvider = new CachedSecretProvider(
+                spyTestProvider,
+                new CacheConfiguration(cacheInterval),
+                new MemoryCache(new MemoryCacheOptions()));
+
+            // Act
+            string actualFirst = await cachedSecretProvider.Get(keyName);
+            await Task.Delay(TimeSpan.Zero);
```
tomkerkhove: Given timespan is zero, do we actually need it?

tomkerkhove: Anyhow, merged it anyway. Feel free to create a new PR if you agree.

stijnmoreels: Ah, could be changed to 100ms or so.

Issue #10, opened 2018-10-08
Provide support for certificate authentication for Azure Key Vault


tomkerkhove:
Provide support for certificate authentication instead of basic authentication.

tomkerkhove:
@ToonVanhoutte feel free to add the configuration suggestions you had

ToonVanhoutte:
This feature will be very helpfull for setting up TLS mutual authentication, where the backend service wants to verify the TLS client certificate (of the API manager). Base requirement would be to validate on the thumbprint, like APIM does: <authentication-certificate thumbprint="thumbprint" />. T

his thumbprint should be configurable, preferably with an option to have it in KeyVault (with caching).

tomkerkhove:
@ToonVanhoutte 	 I think you're more suggesting how https://github.com/arcus-azure/arcus.webapi/issues/31 should work, is that correct?

tomkerkhove:
@arcus-azure/contributors When you would use Certificate based authentication would you prefer to:
1. Let the library handle how it should read the cert from your system?
2. Or give the `X509Certificate2` to the library which takes it from there?

When choosing for 1, where would you store it from which we get it.

ToonVanhoutte:
Option 2 makes sense in this case, imo.

tomkerkhove:
👌, aligns with @stijnmoreels his suggestion.

Will keep it open to gather more feedback.

tomkerkhove:
/remind me in 4 days to proceed based on user feedback

tomkerkhove:
X509Certificate2 it is!

PR #59, opened 2019-05-15
FEAT: certificate based authenticate Key Vault client

stijnmoreels:
Besides client-id/client-secret; we should also provide a way to
authenticate the Key Vault client with client-id/certificate.

 #10

### DIFF
--- a/docs/auth/azure-key-vault.md
+++ b/docs/auth/azure-key-vault.md
@@ -30,4 +30,16 @@ var vaultConfiguration = new KeyVaultConfiguration(keyVaultUri);
 var keyVaultSecretProvider = new KeyVaultSecretProvider(vaultAuthenticator, vaultConfiguration);
 ```
 
+### Certificate Based
+Authentication via client ID and certificate is supported with the `CertifidateBasedAuthentication`.
+
+```csharp
+var clientId = Configuration.GetValue<string>("Arcus:ServicePrincipal:ClientId");
+X509Certificate2 certificate = ...
+
+var vaultAuthenticator = new CertificateBasedAuthentication(clientId, certificate);
+var vaultConfiguration = new KeyVaultConfiguration(keyVaultUri);
+var keyVaultSecretProvider = new KeyVaultSecretProvider(vaultAuthenticator, vaultConfiguration);
+```
+
 [&larr; back](/)
--- a/docs/auth/azure-key-vault.md
+++ b/docs/auth/azure-key-vault.md
@@ -30,7 +30,7 @@ var vaultConfiguration = new KeyVaultConfiguration(keyVaultUri);
 var keyVaultSecretProvider = new KeyVaultSecretProvider(vaultAuthenticator, vaultConfiguration);
 ```
 
-### Certificate Based
+### Certificate
 Authentication via client ID and certificate is supported with the `CertifidateBasedAuthentication`.
 
 ```csharp
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/CertificateBasedAuthentication.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/CertificateBasedAuthentication.cs
@@ -13,20 +13,20 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
     /// </summary>
     public class CertificateBasedAuthentication : IKeyVaultAuthentication
     {
-        private readonly string _applicationId;
+        private readonly string _clientId;
         private readonly X509Certificate2 _certificate;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="CertificateBasedAuthentication"/> class.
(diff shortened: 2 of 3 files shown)


stijnmoreels:
Not sure if this should be included in the integration test? I see that the 'managed service identity' impelemtnation isn't used anywhere?

tomkerkhove:
We should definitely have an integration test for this scenario. The reason why we don't have one for Managed Service Identity is that this is handled via the Azure infrastructure which we cannot use locally unless we use hacks.

Tracking lack of tests for MSI in #60

src/Arcus.Security.Providers.AzureKeyVault/Authentication/CertificateBasedAuthentication.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        private readonly string _applicationId;
+        private readonly X509Certificate2 _certificate;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CertificateBasedAuthentication"/> class.
+        /// </summary>
+        /// <param name="applicationId">The identifier of the application requesting the authentication token.</param>
+        /// <param name="certificate">The certificate that is used as credential.</param>
+        public CertificateBasedAuthentication(string applicationId, X509Certificate2 certificate)
```
tomkerkhove: We use `clientId` in the other authentication option, can you please align?
https://github.com/arcus-azure/arcus.security/blob/master/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs#L25

src/Arcus.Security.Providers.AzureKeyVault/Authentication/CertificateBasedAuthentication.cs
```diff
@@ -0,0 +1,51 @@
...
+            _certificate = certificate;
+        }
+
+        /// <summary>
+        ///     Authenticates with Azure Key Vault
+        /// </summary>
+        /// <returns>A <see cref="IKeyVaultClient" /> client to use for interaction with the vault</returns>
+        public Task<IKeyVaultClient> Authenticate()
+        {
+            IKeyVaultClient client = new KeyVaultClient((async (authority, resource, scope) =>
```
tomkerkhove: We don't tend to do inline stuff like this, please move to a dedicated method similar to our other authentication options: https://github.com/arcus-azure/arcus.security/blob/master/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ServicePrincipalAuthenticator.cs#L54

src/Arcus.Security.Providers.AzureKeyVault/Authentication/CertificateBasedAuthentication.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                var authenticationContext = new AuthenticationContext(authority);
```
tomkerkhove: Please see if we can refactor to share some code with ServicePrincipalAuthenticator

src/Arcus.Security.Providers.AzureKeyVault/Authentication/CertificateBasedAuthentication.cs
stijnmoreels: Azure docs talk about Application ID; but OK to allgin.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/CertificateBasedAuthentication.cs
stijnmoreels: Hmm, only the 

```csharp
var authenticationContext = new AuthenticationContext(authority);
```

is a duplicate. Doesn't seem right to refactor this out.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/CertificateBasedAuthentication.cs
tomkerkhove: Then align the other part ;)

stijnmoreels: :) OK

tomkerkhove: That's a fair point, tracking this in #61 to align this

docs/auth/azure-key-vault.md
```diff
@@ -30,4 +30,16 @@ var vaultConfiguration = new KeyVaultConfiguration(keyVaultUri);
 var keyVaultSecretProvider = new KeyVaultSecretProvider(vaultAuthenticator, vaultConfiguration);
 ```
 
+### Certificate Based
```
tomkerkhove: Maybe just call it `Certificate`?

stijnmoreels: Yes, that'll be better

tomkerkhove:
Thanks!

PR #68, opened 2019-07-17
DOC: running integration tests

### LINKED ISSUE #66: Document how to run integration tests
The Event Grid repository has an introduction on how to run the integration tests present in the repo. This repository hasn't got any such introduction.

Could be reasonable to add this here as well. There are also other environment variables that needs to be configured.

stijnmoreels:
Describe how the integration tests can be run with the ARM template and
environment variables.

Closes #66


stijnmoreels:
Something like this? Not sure but the ARM template doesn't seem to work like the one in the event grid repo.

PR #69, opened 2019-07-17
TEST: stable transistion from 'ClientId' > 'ApplicationId' environment variable

### LINKED ISSUE #61: Use application id instead of client id in Azure AD context
Use application id instead of client id in Azure AD context as this is what Microsoft is using.

We could:
- Perform a breaking change and just update the environment variables
- Provide a smooth transition to support both current and new environment variable and remove the current one when performing breaking change later on.

stijnmoreels:
Azure uses the terminology 'ApplicationId' (as do we in code).
The environment variable, though, doesn't reflect this terminology.

Closes #61


stijnmoreels:
> Feel free to just remove `Arcus:ServicePrincipal:ClientId`

Not very stable change then?

tomkerkhove:
It's our own internal integration testing so I don't really see an issue with that.

What is your concern?

stijnmoreels:
> What is your concern?
That everyone, inclulding us and the build variables, has to be changed for this to work. Every other dev that already has run these tests locally also.

tomkerkhove:
Frankly, I think that's ok - Only you and me work on this and can be a one off switch in the build

PR #71, opened 2019-08-02
FEAT: add 'ISecretProvider' with '.AddAzureKeyVault' configuration builder

stijnmoreels:
Adds an extension to the IConfigurationBuilder so we can pass-in our own
ISecretProvider when replacing configuration tokens with secret values.


tomkerkhove:
Relates to #72

PR #76, opened 2019-08-14
DOC: add missing XML documentation for public members

stijnmoreels:


build/github-package-registry.yml
```diff
@@ -0,0 +1,29 @@
+# Publish project to NuGet
```
tomkerkhove: This is by accident I presume?

build/build.yml
```diff
@@ -0,0 +1,151 @@
+name: Build Arcus.Security CI
```
tomkerkhove: This is by accident I presume?

PR #80, opened 2019-08-20
FEAT: add YAML release to MyGet

stijnmoreels:


stijnmoreels:
> Looks good, can you send a PR after merging to verify that it still works + our bot kicks in?

I guess we have to change something before the bot works?

tomkerkhove:
As we are just using MyGet the bot should work to be honest.

tomkerkhove:
Just had a look, bot filters on *all packages which we do not have for security so won't work here.

Issue #63, opened 2019-06-26
Move to YAML-based builds


tomkerkhove:
Move to YAML based builds which are stored in the GitHub repo under `build` folder.

- [X] Migrate CI Build
- [x] Migrate release to GitHub Package Registry / MyGet.Org
- [x] Migrate release to NuGet.Org
- [x] Migrate creation of GitHub release
- [x] Update CI badge in README

tomkerkhove:
Let's stick with MyGet for now @stijnmoreels, see https://github.com/arcus-azure/arcus/issues/29

PR #81, opened 2019-08-20
DOC: update README.md with YAML build

stijnmoreels:
Relates to #63


tomkerkhove:
See comment in Event Grid Proxy PR

PR #86, opened 2019-08-21
FIX: remove manual variable in MyGet release

stijnmoreels:
When we remove this variable in the variables block in the YAML file, we can set the variable at queue-time.

Relates to #63

### DIFF
--- a/build/myget-release.yml
+++ b/build/myget-release.yml
@@ -11,6 +11,7 @@ variables:
     value: '2.2.105'
   - name: 'Project'
     value: 'Arcus.Security'
+  # 'Package.Version.ManualTrigger' is added as queue-time variable on build in Azure DevOps
 
 stages:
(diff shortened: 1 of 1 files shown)


build/myget-release.yml
```diff
@@ -11,8 +11,6 @@ variables:
     value: '2.2.105'
   - name: 'Project'
     value: 'Arcus.Security'
-  - name: 'Package.Version.ManualTrigger'
```
tomkerkhove: I think ideally we add the variable as a comment to show that it exists, but is configured on the build itself.

We do that for AD secret for example - https://github.com/arcus-azure/arcus.security/blob/master/build/ci-build.yml#L18

stijnmoreels: Ah, OK; good suggestion!

PR #88, opened 2019-08-28
TEST: add integration test for KeyVaultSecretProvider with MSI authentication

### LINKED ISSUE #60: Provide integration test for Managed Service Identity authentication
Provide integration test for Managed Service Identity authentication, if possible.

This depends on the Azure Infrastructure but if I'm not mistaken this can also be "hacked" via an environment variable or this approach - https://stackoverflow.com/questions/47756248/how-to-use-msi-for-development-in-visual-studio

We need to see if this works both locally & CI + if we need docs on running the tests locally.

stijnmoreels:
Closes #60


stijnmoreels:
Hmm, a little help? Not sure if you could just reference another variable while defining a new one. But maybe I'm missing something else.

tomkerkhove:
Variables can be used in others without a problem, I'll have a look when I have some time

build/myget-release.yml
```diff
@@ -11,6 +11,8 @@ variables:
     value: '2.2.105'
   - name: 'Project'
     value: 'Arcus.Security'
+  - name: 'AzureServicesAuthConnectionString'
```
tomkerkhove: Do you need this one perse? I'll add it to variable group for sake of being sure

tomkerkhove:
Is this supported via CLI @stijnmoreels 	?

Looks like it tries the following providers:
> Parameters: Connection String: [No connection string specified], Resource: https://vault.azure.net, Authority: https://login.windows.net/7517bc42-bcf8-4916-a677-b5753051f846. Exception Message: Tried to get token using Managed Service Identity. Access token could not be acquired. MSI ResponseCode: BadRequest, Response: {"error":"invalid_request","error_description":"Identity not found"}
Parameters: Connection String: [No connection string specified], Resource: https://vault.azure.net, Authority: https://login.windows.net/7517bc42-bcf8-4916-a677-b5753051f846. Exception Message: Tried to get token using Visual Studio. Access token could not be acquired. Environment variable LOCALAPPDATA not set.
Parameters: Connection String: [No connection string specified], Resource: https://vault.azure.net, Authority: https://login.windows.net/7517bc42-bcf8-4916-a677-b5753051f846. Exception Message: Tried to get token using Azure CLI. Access token could not be acquired. ERROR: Please run 'az login' to setup account.

Maybe task should be run as part of Azure CLI step, although I'm not sure that iwll help

build/myget-release.yml
stijnmoreels: Yeah, docs says:

> or set an environment variable named **AzureServicesAuthConnectionString** to **RunAs=App;AppId=AppId;TenantId=TenantId;AppKey=Secret**. You need to replace **AppId**, **TenantId**, and **Secret** with actual values.

stijnmoreels:
> Maybe task should be run as part of Azure CLI step, although I'm not sure that iwll help

Hmm, yeah maybe.

build/myget-release.yml
tomkerkhove: 👌 Added to variable group so feel free to remove it here

tomkerkhove:
Would be good to use our new approach to test this: https://github.com/arcus-azure/arcus.security/pull/90

Should give the same result as we don't have to test the magic scenario perse imo

stijnmoreels:
> Would be good to use our new approach to test this: #90
> 
> Should give the same result as we don't have to test the magic scenario perse imo

Ok, I updated the integration test with that; and it seems to work! Cool!

tomkerkhove:
Awesome, thanks!

Issue #89, opened 2019-09-04
Support AzureServiceTokenProvider constructor optional parameters


BlackstarSolar:
While setting the `AzureServiceTokenProvider` connection string via the _AzureServicesAuthConnectionString_ environment variable works in many cases, in some cases (e.g. Azure App Service configuration) building a connection string, as an environment variable, often from other environment variables isn't practical/possible.

Additionally specifying the `azureAdInstance` parameter (for when not running in Azure public cloud) via environment variable is not currently supported.

Supporting the optional `connectionString` and `azureAdInstance` parameters would provide greater flexibility.

tomkerkhove:
That's a good suggestion, thanks!

@stijnmoreels 	This might help you with the issues you are suffering from now as well.

tomkerkhove:
Thank you @BlackstarSolar 	!

PR #90, opened 2019-09-04
Added support for connection string and azure AD instance for Managed Identity auth

BlackstarSolar:
Addresses arcus-azure/arcus.security#89

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
@@ -21,7 +21,7 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
         /// Initializes a new instance of the <see cref="ManagedServiceIdentityAuthenticator"/> class.
         /// </summary>
         /// <param name="connectionString">The connection string to use to authenticate.</param>
-        /// <param name="azureAdInstance">The azure AD instance to use to authenticate.</param>
+        /// <param name="azureAdInstance">The azure AD instance to use to authenticate, if applicable.</param>
         public ManagedServiceIdentityAuthenticator(string connectionString = null, string azureAdInstance = null)
         {
             _connectionString = connectionString;
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
@@ -27,7 +27,7 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
         /// <summary>
         /// Initializes a new instance of the <see cref="ManagedServiceIdentityAuthenticator"/> class.
         /// </summary>
-        /// <param name="connectionString">The connection string to use to authenticate.</param>
+        /// <param name="connectionString">The connection string to use to authenticate, if applicable.</param>
         /// <param name="azureAdInstance">The azure AD instance to use to authenticate, if applicable.</param>
         public ManagedServiceIdentityAuthenticator(string connectionString = null, string azureAdInstance = null)
         {
--- a/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
@@ -13,6 +13,21 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
     public class ManagedServiceIdentityAuthenticator : IKeyVaultAuthentication, IKeyVaultAuthenticator
 #pragma warning restore 618
     {
+        private readonly string _connectionString;
+
+        private readonly string _azureAdInstance;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ManagedServiceIdentityAuthenticator"/> class.
+        /// </summary>
+        /// <param name="connectionString">The connection string to use to authenticate.</param>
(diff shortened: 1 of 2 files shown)


src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
```diff
@@ -13,6 +13,21 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
...
+        private readonly string _connectionString;
+
+        private readonly string _azureAdInstance;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ManagedServiceIdentityAuthenticator"/> class.
+        /// </summary>
+        /// <param name="connectionString">The connection string to use to authenticate.</param>
+        /// <param name="azureAdInstance">The azure AD instance to use to authenticate.</param>
+        public ManagedServiceIdentityAuthenticator(string connectionString = null, string azureAdInstance = null)
```
tomkerkhove: Frankly I think we should provide 3 overloads:
1. Constructor without any parameters
2. One with connection string only
3. One with both of these

This would allow us to guard on the values and enforce providing the required values. Next to that, it would be clearer for consumers when to use what.

Do you agree? How about you @stijnmoreels 	?

tomkerkhove:
Thank you for your contribution btw @BlackstarSolar!

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
BlackstarSolar: The parameters are optional so we already have the 3 overloads you suggest.  This mirror (and slightly improves) the [`AzureServiceTokenProvider` constructor](https://github.com/Azure/azure-sdk-for-net/blob/ddda7cb74b979f03bb03e240c06c924914ee8bdd/src/SdkCommon/AppAuthentication/Azure.Services.AppAuthentication/AzureServiceTokenProvider.cs#L61)

There wouldn't be any guarding required on `connectionString`.  `null` is a valid value and causes the `AzureServiceTokenProvider` to fall back to default "magic" of looking at the `AzureServicesAuthConnectionString`, or using System Assigned Identity, VS token provider or Azure CLI token provider.

The guarding is already in place for azureAdInstance ([see comment on line 65](https://github.com/arcus-azure/arcus.security/pull/90/files/fa5e24d73361a0da9791d418622ac1390c266385#diff-e1f5781b0670280ed5307045e4c9767fR65))

BlackstarSolar:
> Would you be up for updating the documentation as well?

Sure I can take a look.  Assume I only need to update the .md not the .html too?

> Thank you for your contribution btw @BlackstarSolar!

You're welcome :-)

tomkerkhove:
> > Would you be up for updating the documentation as well?
> 
> Sure I can take a look. Assume I only need to update the .md not the .html too?

Jup that is correct - Thanks!

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
tomkerkhove: Yes and no, while I agree I think it would be good to have an empty controller next to this one at least because it might confuse people who are new to Managed Identity or do not have the technical background on how it works.

Managed Identity does "magic" for 90% of the use-cases and would prefer to keep it like that if it's ok for you?

tomkerkhove: I agree on the other aspects concerning guarding, that's ok for me.

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
```diff
@@ -13,6 +13,21 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
...
     {
+        private readonly string _connectionString;
+
+        private readonly string _azureAdInstance;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ManagedServiceIdentityAuthenticator"/> class.
+        /// </summary>
+        /// <param name="connectionString">The connection string to use to authenticate.</param>
+        /// <param name="azureAdInstance">The azure AD instance to use to authenticate.</param>
```
tomkerkhove:
```suggestion
        /// <param name="azureAdInstance">The azure AD instance to use to authenticate, if applicable.</param>
```

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
```diff
@@ -13,6 +13,21 @@ namespace Arcus.Security.Providers.AzureKeyVault.Authentication
...
 #pragma warning restore 618
     {
+        private readonly string _connectionString;
+
+        private readonly string _azureAdInstance;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ManagedServiceIdentityAuthenticator"/> class.
+        /// </summary>
+        /// <param name="connectionString">The connection string to use to authenticate.</param>
```
tomkerkhove:
```suggestion
        /// <param name="connectionString">The connection string to use to authenticate, if applicable.</param>
```

src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthenticator.cs
BlackstarSolar: OK, I'll add a parameterless constructor when I get into the office tomorrow morning and push an update.

BlackstarSolar: Done.

PR #91, opened 2019-09-06
TEST: add integration test with MSI using default config

stijnmoreels:
Uses the environment-variable-approach to test the default config of MSI authentication.

Relates to #60


stijnmoreels:
> Some good thinking over here, thanks man!

:) thanks, if we find other places where we need to dispose variables. I suggest we create an disposable type where we set/remove the variable.
But for now try/finally will do.

Issue #93, opened 2019-09-25
ISecretProvider documentation is outdated


fgheysels:
**Describe the bug**
The code sample in the [documentation page](https://security.arcus-azure.net/features/key-vault/extensions/iconfiguration-integration) refers to an obsolete constructor.

**To Reproduce**
When using the code from the documentation, you'll get a warning that says that the constructor for `KeyVaultSecretProvider` is obsolete:

```csharp
var vaultAuthenticator = new ManagedServiceIdentityAuthenticator();
var vaultConfiguration = new KeyVaultConfiguration(keyVaultBaseUrl);
var yourSecretProvider = new KeyVaultSecretProvider(vaultAuthenticator, vaultConfiguration);
```

>The call is ambiguous between the following methods or properties: 'KeyVaultSecretProvider.KeyVaultSecretProvider(IKeyVaultAuthenticator, IKeyVaultConfiguration)' and 'KeyVaultSecretProvider.KeyVaultSecretProvider(IKeyVaultAuthentication, IKeyVaultConfiguration)'	

**Expected behavior**
Reference documentation that uses the latest version of the library :)

**Additional context**
 - Version: 0.4.0

tomkerkhove:
Clarified in #96, thank you!

PR #96, opened 2019-09-25
FIX: update SecretProvider documentation to not use obsolete types

fgheysels:
- Updated the documentation to use the `IKeyVaultAuthentication` type instead of the `IKeyVaultAuthenticator`
- Added some extra clarification regarding caching

#93


tomkerkhove:
Thank you!

PR #98, opened 2019-09-27
Use YAML template from arcus-azure/azure-devops-templates to push preview package

tomkerkhove:
Use YAML template from arcus-azure/azure-devops-templates to push preview package

### DIFF
--- a/build/myget-release.yml
+++ b/build/myget-release.yml
@@ -7,6 +7,7 @@ resources:
   - repository: templates
     type: github
     name: arcus-azure/azure-devops-templates
+    endpoint: arcus-azure
 
 variables:
   - group: 'Arcus Security - Integration Testing'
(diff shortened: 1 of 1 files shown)


build/myget-release.yml
```diff
@@ -153,10 +160,4 @@ stages:
             inputs:
               artifact: 'Build'
               path: '$(Build.SourcesDirectory)'
-          - task: NuGetCommand@2
-            displayName: 'Push to MyGet.org'
-            inputs:
-              command: push
-              packagesToPush: 'src/**/*.nupkg'
```
stijnmoreels: Would this not be a argument that we should pass in?

tomkerkhove: It's defaulting to that, but you can override if there is a need.

Can add it for sake of being explicit if you want

stijnmoreels: Ah, ok, nevermind then.

PR #99, opened 2019-09-27
Refactor build to use YAML templates

tomkerkhove:
Refactor build to use YAML templates


build/myget-release.yml
```diff
@@ -160,4 +102,4 @@ stages:
             inputs:
               artifact: 'Build'
               path: '$(Build.SourcesDirectory)'
-          - template: build/nuget/publish-preview-package.yml@templates
```
tomkerkhove: I'll delete the old file once merged.

build/ci-build.yml
```diff
@@ -67,18 +75,10 @@ stages:
...
-          - task: DotNetCoreInstaller@0
-            displayName: 'Import .NET Core SDK ($(DotNet.Sdk.Version))'
-            inputs:
-              version: '$(DotNet.Sdk.Version)'
-          - task: DotNetCoreCLI@2
-            displayName: 'Run unit tests'
-            inputs:
-              command: test
-              projects: 'src/**/$(Project).Tests.Unit.csproj'
-              arguments: '--configuration $(Build.Configuration)'
```
tomkerkhove: Ha no version prefix.

What would you do @stijnmoreels : Do we add something fixed and don't care or see if we can add ifs?

stijnmoreels: Version of the build nuget package? Or from the .net SDK? I think you talk about the variable that has 'preview' as value?

tomkerkhove: Other builds use --version-prefix, which is now part of the task but not the CI build.

stijnmoreels: Yeah, could we have optional parameters in the template? So we can ignore the version prefix in the CI build?

tomkerkhove: We could, we'd just need to figure out how it works in terms of passing the parameter or not.

Should have some time later on to figure that out unless you can already do some research on this ofcourse.

tomkerkhove: Resolving as we're tracking this in https://github.com/arcus-azure/azure-devops-templates/issues/7

PR #101, opened 2019-10-16
CHORE: merge MyGet release into CI build

stijnmoreels:
We don't need a seperate duplicate of the build to release to MyGet.

 Relates to #63


build/ci-build.yml
```diff
@@ -3,16 +3,23 @@ name: $(date:yyyyMMdd)$(rev:.r)
 trigger:
   branches:
     include:
-    - master
+      - master
```
tomkerkhove: Is this on purpose? I don't think we need this, do we? 🤔
https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=example#triggers

build/ci-build.yml
```diff
@@ -3,16 +3,23 @@ name: $(date:yyyyMMdd)$(rev:.r)
...
     include:
-    - src/*
+      - src/*
 
 pr:
   paths:
     include:
-    - src/*
-    - build/ci-build.yml
+      - src/*
```
tomkerkhove: Same here

build/ci-build.yml
```diff
@@ -3,16 +3,23 @@ name: $(date:yyyyMMdd)$(rev:.r)
 trigger:
   branches:
     include:
-    - master
+      - master
   paths:
     include:
-    - src/*
+      - src/*
```
tomkerkhove: Same here

build/ci-build.yml
stijnmoreels: Ah, no; I don't think so. We had this before on all CI builds; but as it turns out, this is triggered by default.

tomkerkhove: I think we actually need them but I was more referring to the additional spaces

build/ci-build.yml
```diff
@@ -32,6 +28,35 @@ stages:
         pool:
           vmImage: 'ubuntu-16.04'
         steps:
+          - powershell: |
```
tomkerkhove: Can we use this task? https://github.com/arcus-azure/azure-devops-templates/blob/master/nuget/determine-pr-version.yml

build/ci-build.yml
```diff
@@ -1,18 +1,13 @@
 name: $(date:yyyyMMdd)$(rev:.r)
 
-trigger:
```
tomkerkhove: Let's re-add this

PR #109, opened 2019-11-04
Use fluent approach in example when adding KeyVault ConfigProvider

fgheysels:
To make it more obvious that there exists a more fluent way to use the caching capabilities, I've modified the documentation on how to integrate the arcus KeyVault Provider with the ASP.NET configuration.


tomkerkhove:
LGTM, thanks!

PR #116, opened 2019-11-29
CHORE: use build YAML template with version property

stijnmoreels:
Uses the YAML build template with the version property.

Relates to [#7](https://github.com/arcus-azure/azure-devops-templates/issues/7)


stijnmoreels:
Ok, but a bit besides the point of this series of PR's.
Comes later when we have agreed on a approach.

tomkerkhove:
https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#matrix
This actually sounds like what we want, no? 🤔

PR #120, opened 2020-01-07
FEAT: invalidated cached secret

stijnmoreels:
Relates to https://github.com/arcus-azure/arcus.webapi/issues/86#issuecomment-558220246


docs/features/secrets/general.md
```diff
@@ -60,4 +60,13 @@ After a hard refresh you can use the latest secret again and proceed your work.
 Secret secret = await cachedSecretProvider.GetSecretAsync("EventGrid-AuthKey", ignoreCache: true);
 ```
 
+## Invalidates a secret from the cach
```
tomkerkhove:
```suggestion
## Invalidates a secret from the cache
```

docs/features/secrets/general.md
```diff
@@ -60,4 +60,13 @@ After a hard refresh you can use the latest secret again and proceed your work.
...
+In some scenarios you'd like to remove the entry from the cache, so the next the secret is retrieved, a new lookup will be done in the provider.
```
tomkerkhove: How about this?

```suggestion
In some scenarios you'd like to remove a cache entry so that the secret will be retrieved from the provider when a new lookup will be done.
```

docs/features/secrets/general.md
```diff
@@ -60,4 +60,13 @@ After a hard refresh you can use the latest secret again and proceed your work.
...
+
+This is useful for scenario's where the secret is updated and you need to tell the cache somehow.
+
+```csharp
+await cachedSecretProvider.InvalidateSecretAsync("EventGrid-AuthKey");
```
tomkerkhove: `InvalidateSecretAsync` or `InvalidateAsync`? But I presume we have `GetSecretAsync`?

```suggestion
await cachedSecretProvider.InvalidateAsync("EventGrid-AuthKey");
```

src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
```diff
@@ -129,5 +130,16 @@ public async Task<Secret> GetSecretAsync(string secretName, bool ignoreCache)
...
         }
+
+        /// <summary>
+        /// Removes the secret with the given <paramref name="secretName"/> from the cache;
+        /// so the next time <see cref="GetSecretAsync(string)"/> is called, a new version of the secret will be added back to the cache.
+        /// </summary>
+        /// <param name="secretName">The name of the secret that should be removed from the cache.</param>
+        public Task InvalidateSecretAsync(string secretName)
+        {
+            _memoryCache.Remove(secretName);
```
tomkerkhove: What if it's not there? Does it throw? Should we check?

src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
```diff
@@ -129,5 +130,16 @@ public async Task<Secret> GetSecretAsync(string secretName, bool ignoreCache)
...
             return secret;
         }
+
+        /// <summary>
+        /// Removes the secret with the given <paramref name="secretName"/> from the cache;
+        /// so the next time <see cref="GetSecretAsync(string)"/> is called, a new version of the secret will be added back to the cache.
+        /// </summary>
+        /// <param name="secretName">The name of the secret that should be removed from the cache.</param>
+        public Task InvalidateSecretAsync(string secretName)
+        {
```
tomkerkhove: Should we guard for empty `secretName`?

docs/features/secrets/general.md
tomkerkhove: It's `CachedSecretProvider.GetSecretAsync(string)` so it's ok.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
```diff
@@ -247,5 +247,35 @@ public async Task CachedSecretProvider_GetTwoSecretsAsync_SkippedCache_ShouldRet
...
+            const string keyName = "MySecret";
+            var cachedSecretProvider = new CachedSecretProvider(
+                testProviderStub,
+                new CacheConfiguration(cacheInterval));
+
+            Secret actualFirst = await cachedSecretProvider.GetSecretAsync(keyName);
+            testProviderStub.SecretValue = expectedSecondSecret;
+
+            // Act
+            await cachedSecretProvider.InvalidateSecretAsync(keyName);
```
tomkerkhove: Should we do a new get secret to see if it first caches it correct, or do you rely on the other tests for that?

docs/features/secrets/general.md
stijnmoreels: Yeah, that's why I added the `...Secret...`. part

src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
stijnmoreels: Ah, yes; forgot.

src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
stijnmoreels: `null` will be returned for `MemoryCache`; but it's an interface we take in the ctor; so could be anything.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: Yeah, I thought of other scenarios; but I guess all those are already covered.

tomkerkhove: Are they already covered or should we still add it? Just want to make sure it's covered 😅

src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
tomkerkhove: Hm there is no way to verify if it's there so it's ok.

src/Arcus.Security.Tests.Unit/Core/CachedSecretProviderTests.cs
stijnmoreels: I think that we have all the useful scenario's covered.

PR #124, opened 2020-03-03
Chore - update Security projects with netcore3.0

stijnmoreels:
Relates to https://github.com/arcus-azure/arcus/issues/50


fgheysels:
Just an additional question:  In the `TargetFrameworks` property, netcoreapp3.0 is mentionned.  However, what about .NET Core 3.1, 3.x ?

tomkerkhove:
Good point, it should be 3.1 indeed as 3.0 is already EOL

PR #129, opened 2020-05-01
Feature - add expiration date to secret

### LINKED ISSUE #128: Surface expiration date information from Azure Key Vault
Surface expiration date information from Azure Key Vault so that:
- People know when it will expire
- Our caching can cache it accordingly

However, this field should be optional in `Secret` class since not every provider supports this AFAIK.

stijnmoreels:
Closes #128


src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -84,7 +84,10 @@ public async Task<Secret> GetSecretAsync(string secretName)
                     return null;
                 }
 
-                return new Secret(secretBundle.Value, secretBundle.SecretIdentifier?.Version);
+                return new Secret(
```
stijnmoreels, line 87: The Key Vault `SecretBundle` works with `DateTime` instead of `DateTimeOffset`.

src/Arcus.Security.Core/Secret.cs
```diff
@@ -34,5 +36,10 @@ public Secret(string value, string version)
         /// </summary>
         /// <remarks>Version is not checked for <c>null</c>.</remarks>
         public string Version { get; }
+
+        /// <summary>
+        /// Gets the optional expiration date of the secret.
```
tomkerkhove, line 41: Since it's nullable it's clear that it's optional, no?
```suggestion
        /// Gets the expiration date of the secret.
```

src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: Looks like it's in UTC so I would map it to a `DateTimeOffset` for others
https://docs.microsoft.com/en-us/cli/azure/keyvault/secret?view=azure-cli-latest#az-keyvault-secret-set-attributes

stijnmoreels: Ok, done.

PR #140, opened 2020-05-25
Update consume-from-key-vault.md

SamVanhoutte:
Updated missing link in the docs

### DIFF
--- a/docs/features/secrets/consume-from-key-vault.md
+++ b/docs/features/secrets/consume-from-key-vault.md
@@ -12,6 +12,6 @@ var vaultConfiguration = new KeyVaultConfiguration(keyVaultUri);
 var keyVaultSecretProvider = new KeyVaultSecretProvider(vaultAuthentication, vaultConfiguration)
 ```
 
-You can find a list of supported authentication schemes for Azure Key Vault [here](./../../features/auth/azure-key-vault.md).
+You can find a list of supported authentication schemes for Azure Key Vault [here](./../../features/auth/azure-key-vault).
 
 [&larr; back](/)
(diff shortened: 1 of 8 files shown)


docs/features/secrets/consume-from-key-vault.md
```diff
@@ -12,6 +12,6 @@ var vaultConfiguration = new KeyVaultConfiguration(keyVaultUri);
 var keyVaultSecretProvider = new KeyVaultSecretProvider(vaultAuthentication, vaultConfiguration)
 ```
 
-You can find a list of supported authentication schemes for Azure Key Vault [here](./../../auth/azure-key-vault).
+You can find a list of supported authentication schemes for Azure Key Vault [here](./../../features/auth/azure-key-vault.md).
```
stijnmoreels, line 15: No need for the `.md` extension, the Jekyll docs will use that directly.

docs/features/secrets/consume-from-key-vault.md
```diff
@@ -7,7 +7,7 @@ layout: default
 You can easily create a Key Vault secret provider - The only thing you need to do is specify how you want to configure and to what vault.
 
 ```csharp
-var vaultAuthentication = new ManagedServiceIdentityAuthentication();
+var vaultAuthentication = new ManagedServiceIdentityAuthenticator();
```
stijnmoreels, line 10: In older versions, this was indeed the case, (see older docs) but this is the current docs of the most recently released version, where this class was renamed. See [this](https://github.com/arcus-azure/arcus.security/blob/master/src/Arcus.Security.Providers.AzureKeyVault/Authentication/ManagedServiceIdentityAuthentication.cs)

SamVanhoutte: okay - I had to upgrade to the preview version for this.  will roll back the above change

PR #154, opened 2020-07-02
Feature - provide secret store for Azure Functions

### LINKED ISSUE #153: Provide support for Secret Store in Azure Functions
Provide support for Secret Store in Azure Functions by providing our extensions on the IFunctionHostBuilder.

stijnmoreels:
Closes #153


stijnmoreels:
Voila, nailed it! Successfully created an Azure Function and used the MyGet package of the Azure Functions project here created. A secret provider can be added by the created extension here and is successfully injected in the created function's ctor.
🕺 .

src/Arcus.Security.AzureFunctions/Arcus.Security.AzureFunctions.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <Authors>Arcus</Authors>
+    <Company>Arcus</Company>
+    <Description>Contains Azure Functions core functionality for Arcus.Security</Description>
```
tomkerkhove, line 7: Maybe better since it's a phrase?
```suggestion
    <Description>Contains Azure Functions core functionality for Arcus Security</Description>
```

src/Arcus.Security.Core/InternalsVisibleTo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Arcus.Security.AzureFunctions")]
```
tomkerkhove, line 3: What does this do? 🤔

docs/preview/features/secret-store/azure-functions.md
```diff
@@ -0,0 +1,74 @@
+---
```
tomkerkhove, line 1: Just a thought but maybe we should just move it to the other doc as well in a dedicated header similar to `Using secret store with Azure Functions`?

Because the only things that are different here are:
1. NuGet 📦 
2. `Startup` sample
3. Consumption example

All the rest is the same or not?

docs/preview/features/secret-store/index.md
```diff
@@ -10,6 +10,8 @@ We provide an approach similar to how `IConfiguration` is built, but with a focu
 
 Once register, you can fetch all secrets by using `ISecretProvider` which will get secrets from all the different registered secret providers.
 
+> :bulb: See [this page](./../../features/secret-store/azure-functions) if you want to use the secret store functionality whitin Azure Functions.
```
tomkerkhove, line 13: If we use my suggestion to merge the function page in this one, I would keep this remark as it would help people find the docs below. Thoughts?

src/Arcus.Security.Core/InternalsVisibleTo.cs
stijnmoreels: Yes, was not entirely sure if this is the correct way. I need the `SecretStoreBuilder.RegisterProviders()` internal method in the extension written in the Azure Functions project. This was made internal bc it isn't recommanded to expose it IMHO. But that leaves us with a problem here.
We could introduce a separate model where we call this method, and make that one public, or use `InternalsVisibleTo`. Or another proposal from you?

docs/preview/features/secret-store/azure-functions.md
stijnmoreels: Yes, all the same; it's maybe easier to point to a dedicated page for this. IDK, I was thinking of other environments too where we have to create a dedicated extension for it.
But otherwise, yeah, I can agree on this.

tomkerkhove: Let's start with all in one place because it's easier to split later on than merge. I want to avoid the "spread of information"

stijnmoreels: Yes, done that.

docs/preview/features/secret-store/index.md
stijnmoreels: Yes, done that.

docs/preview/index.md
```diff
@@ -25,6 +25,7 @@ PM > Install-Package Arcus.Security.Providers.AzureKeyVault
     - [Configuration](features/secret-store/provider/configuration)
     - [Environment variables](features/secret-store/provider/environment-variables)
   - [Creating your own secret provider](features/secret-store/create-new-secret-provider)
+  - [For Azure Functions](features/secret-store/azure-functions)
```
tomkerkhove, line 28: This can be removed I think?
```suggestion
```

PR #156, opened 2020-07-06
Reshuffling of documentation in more logical ordering

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

### DIFF
--- a/docs/preview/features/secret-store/index.md
+++ b/docs/preview/features/secret-store/index.md
@@ -12,16 +12,23 @@ Once register, you can fetch all secrets by using `ISecretProvider` which will g
 
 > :bulb: See [this section](#using-secret-store-within-azure-functions) if you want to use the secret store functionality whitin Azure Functions.
 
-## Using secret store within .NET Core (web, worker...) project
+## Built-in secret providers
+Several built in secret providers available in the package.
 
-### Installation
+* [Environment variables](./../../features/secret-store/provider/environment-variables)
+* [Configuration](./../../features/secret-store/provider/configuration)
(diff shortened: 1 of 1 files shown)


docs/preview/features/secret-store/index.md
```diff
@@ -12,16 +12,23 @@ Once register, you can fetch all secrets by using `ISecretProvider` which will g
 
 > :bulb: See [this section](#using-secret-store-within-azure-functions) if you want to use the secret store functionality whitin Azure Functions.
 
-## Using secret store within .NET Core (web, worker...) project
+## Built-in secret providers
```
tomkerkhove, line 15: I've moved this up so that we first show what we support rather than how to use it without knowing if you'll use it

docs/preview/features/secret-store/index.md
```diff
@@ -12,16 +12,23 @@ Once register, you can fetch all secrets by using `ISecretProvider` which will g
...
+Several built in secret providers available in the package.
 
-### Installation
+* [Environment variables](./../../features/secret-store/provider/environment-variables)
+* [Configuration](./../../features/secret-store/provider/configuration)
+* [Azure key vault](./../../features/secret-store/provider/key-vault)
+
+If you require an additional secret providers that aren't available here, please [this document](./../../features/secret-store/create-new-secret-provider) that describes how you can create your own secret provider.
+
+## Installation
```
tomkerkhove, line 24: I've moved this back and removed the header as Functions should be an exception and otherwise we have to update every system we support

PR #164, opened 2020-08-10
Feature - mutate secret names before looking up

### LINKED ISSUE #146: Provide capability to mutate secret names before looking them up
Provide the capability to mutate secret names before looking them up.

This is important because now we have multiple providers and some might use a different approach.

For example, with Azure Key Vault we tend to use `API-Key` while our environment variables use `API_URL` notations. If we were to look for a secret and we provide `API-Key` it should lookup `API_Key`.

Be able to opt-in to influence the secret names by providing a lambda.

Here is an example for environment variables:
```
builder.AddEnvironmentVariables(originalSecretName => originalSecretName.Replace("-", "_"));
```

But ideally every provider that we have now supports this because IConfiguration should use `api:key` instead of `API-KEY`

**Describe alternatives you've considered**
Every provider does this for you but this is more tricky since the convention depends on the customer. A "seperator" could be configured like this:
```
builder.AddEnvironmentVariables(keySeperator: "-");
```

But I'm not sure if that's what we prefer.

stijnmoreels:
Added an `ISecretProvider` proxy during registration of the secret provider source to handle possible secret name mutations.
Also updated existing registration extensions:
- Added possibility to this mutation to the `IConfiguration` registration
- Added possibility to this mutation to the environment variables registration

Provided some (preview) docs in both concrete cases and in the description on 'How to create a custom secret provider?".

Closes #146


stijnmoreels:
Didn't update (just yet) the Azure Key Vault and User Secrets registrations bc I didn't know yet if these should have these?

tomkerkhove:
> Didn't update (just yet) the Azure Key Vault and User Secrets registrations bc I didn't know yet if these should have these?

Yes, please add it there as well!

docs/preview/features/secret-store/provider/configuration.md
```diff
@@ -32,7 +32,11 @@ public class Program
                    .ConfigureSecretStore((HostBuilderContext context, IConfiguration config, SecretStoreBuilder builder) =>
                    {
 #if DEBUG
+                       // Uses the built `IConfiguration` as a secret provider.
                        builder.AddConfiguration(config);
+
+                       // Uses the built `IConfiguration` as secret provider, using `:` instead of `.` when looking up secrets.
+                       builder.AddConfiguration(config, secretName => secretName.Replace(".", ":"));
```
tomkerkhove, line 39: Good example! ❤️ Maybe add an example?
```suggestion
                       // Example - When looking up Queue.Name it will be changed to queue:name
                       builder.AddConfiguration(config, secretName => secretName.Replace(".", ":").ToLower());
```

docs/preview/features/secret-store/provider/environment-variables.md
```diff
@@ -33,6 +33,9 @@ public class Program
 
                        // Uses the environment variables starting with 'ARCUS_' from the environment block associated with the current process.
                        builder.AddEnvironmentVariables(prefix: "ARCUS_");
+
+                       // Uses the environment variables, using underscores and capitals for secret name structure.
+                       builder.AddEnvironmentVariables(mutateSecretName: name => name.Replace(".", "_").ToUpper());
```
tomkerkhove, line 38: Good example! Maybe even extend it to include a prefix:
```suggestion
                       // Example - When looking up Queue.Name it will be changed to ARCUS_QUEUE_NAME
                       builder.AddEnvironmentVariables(mutateSecretName: name => $"ARCUS_{name.Replace(".", "_").ToUpper()}");
```

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <returns>Returns the secret key.</returns>
+        /// <exception cref="ArgumentException">The <paramref name="secretName"/> must not be empty</exception>
+        /// <exception cref="ArgumentNullException">The <paramref name="secretName"/> must not be null</exception>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<string> GetRawSecretAsync(string secretName)
+        {
+            secretName = _mutateSecretName(secretName);
+            Task<string> rawSecretAsync = _implementation.GetRawSecretAsync(secretName);
+
+            if (rawSecretAsync is null)
```
tomkerkhove, line 43: This will only be bubbled if no providers in the store find a match, right?

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                throw new SecretNotFoundException(secretName);
```
tomkerkhove, line 45: Thinking out loud - Should it have the specified name by the user, the mutated one or both?

docs/preview/features/secret-store/create-new-secret-provider.md
```diff
@@ -92,32 +92,51 @@ This section describes how a new secret store source can be added to the pipelin
...
+
+### Adding secret name mutation before looking up secret
+
+When you want secret names 'changed' or 'mutated' before they go through your secret provider (ex. changing `Arcus.Secret` to `ARCUS_SECRET`);
+you can pass allong a custom mutation function during the registration:
+
+```csharp
+public static class SecretBuilderExtensions
+{
+    public static SecretStoreBuilder AddRegistry(this SecretStoreBuilder builder)
```
tomkerkhove, line 132: Or allow users to specify this

docs/preview/features/secret-store/create-new-secret-provider.md
```diff
@@ -92,32 +92,51 @@ This section describes how a new secret store source can be added to the pipelin
    }
```
tomkerkhove, line 92: Let's add a TOC at the top of this article to help people get started

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
stijnmoreels: No, it's actually my paranoia 😅. If people implement the asynchronous method wrong and return `null` instead of a `Task` it will throw a `NullRerenceException` if we try to `await` it.

Like:
```csharp
public Task GetRawSecretAsync(string secretName)
{
    return null;
}
```

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
stijnmoreels: Added descriptions to secret providers that the composite can use when throwing + added the readable transformation to the inner exception message.

docs/preview/features/secret-store/create-new-secret-provider.md
stijnmoreels: Added an example for this.

src/Arcus.Security.Core/CompositeSecretProvider.cs
```diff
@@ -160,32 +162,67 @@ public async Task InvalidateSecretAsync(string secretName)
                 throw new SecretNotFoundException(secretName, noRegisteredException);
             }
 
+            var results = new Collection<(string description, string key, string outcome)>();
```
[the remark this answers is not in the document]
stijnmoreels: 😆 something like that.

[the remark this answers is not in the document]
stijnmoreels: Yes, I added this bc when you add the secret name mutation to the secret store, your 'concrete' secret provider will get wrapped inside a `MutatedSecretNameSecretProvider` (much like with caching, and the `CachedSecretProvider`).

If we then throw an exception _"Can't find secret with name "bar" in provider "{Provider""_, we will have the `MutatedSecretNameSecretProvider` in the logging; making it (IMHO) somewhat useless bc you can't see the _original_ provider.

So, I created this interface `ISecretProviderDescription` that _any_ `ISecretProvider` can implement (ex. `KeyVaultSecretProvider` had `Azure Key Vault`). When I then use secret mutations and the Azure Key Vault secret provider, I can create this exception when the secret is not found:
`> Mutated + Azure Key Vault ('Arcus:Secret:Name' -> 'Arcus.Secret.Name') = NotFound`.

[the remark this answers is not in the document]
stijnmoreels: Some temp models, but I can create something if you can agree on the changes made now and I don't have to revert something.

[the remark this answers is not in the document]
tomkerkhove: I think it's a bit of overkill to be honest, can't we just use the original & mutated name somehow? I don't see how the current exception message is providing a lot of value, to be honest.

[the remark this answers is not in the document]
stijnmoreels: Ok, I'll revert some, it was mostly bc you could see each secret store registration and could see exact which providers failed or couldn't find the secret + the mutation/cached composition.
But I'll revert some so you can only see the secret name mutation.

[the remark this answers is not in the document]
stijnmoreels: Hmm, question, and I think that's the reason for this change: 
_why_ I made this was for writting the output of multiple providers.

So, let's say you have two providers with both mutations... and they both fail to retrieve the secret. Which one would you log? And if a provider without a mutation is added after those two and also fails, which secret name should then be added to the exception?
I think that's why I made this table of contents in the exception message so you can see all of it.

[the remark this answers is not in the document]
tomkerkhove: Frankly, what we need here is logging inside the provider as a verbose trace for the metric that was not found.

The aggregate provider just throws secret not found with the original one and be done with it, no?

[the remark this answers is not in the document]
stijnmoreels: Hmm, no bc we should check all sub-providers in the composite before we can say that the secret was not found. What if the first one throws not found and the second one finds the secret? Then you will want the secret. But if both the first one and the second one throws... That's why this setup. Open for alternatives of course.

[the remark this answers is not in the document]
stijnmoreels: > Frankly, what we need here is logging inside the provider as a verbose trace for the metric that was not found.
> 
> The aggregate provider just throws secret not found with the original one and be done with it, no?

We already traced this, and throws the orignal secret name, it was just by your suggestion that we would maybe need the mutated secret name too that I changed this.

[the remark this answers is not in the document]
tomkerkhove: Then I think it's ok because of the reasons you've mentioned above. We already provide insights on what was not found per provider (trace with mutated name) and throw an exception with original name.

If that is correct, then we can roll this change back - So sorry about this but didn't take this into account :(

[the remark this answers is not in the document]
stijnmoreels: No problem, that's what PR comments are for, I'll update this after I've done with something else.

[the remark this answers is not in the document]
stijnmoreels: Ok, that's done, and I added some examples with the other providers as well (user secrets and key vault).

src/Arcus.Security.Core/SecretStoreBuilder.cs
```diff
@@ -40,19 +42,35 @@ public SecretStoreBuilder(IServiceCollection services)
         /// <remarks>
         ///     The series of secret stores is directly publicly available including the operations so future (consumer) extensions can easily manipulate this series during build-up.
         /// </remarks>
-        public IList<SecretStoreSource> SecretStoreSources { get; } = new List<SecretStoreSource>();
+        public IList<Func<IServiceProvider, SecretStoreSource>> CreateSecretStoreSources { get; } = new List<Func<IServiceProvider, SecretStoreSource>>();
```
stijnmoreels, line 45: Had to change this to have access to the `IServiceProvider` when initiating the `MutatedSecretNameSecretProvider` instance. Technically, this is a breaking change bc it's public, but not sure if it was needed to be public in the first place plus it was/will only be used when you want to manipulated already registered secret sources (so we may not allow that anyway).

stijnmoreels: So, I could make this backwards compatible with obsolete properties or I could leave it as it is.

stijnmoreels: Ah, I could just extend the `SecretSource` itself. hang on...

stijnmoreels: Ok, nevermind this. Changed it.

src/Arcus.Security.Core/CompositeSecretProvider.cs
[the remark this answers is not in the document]
stijnmoreels: Not relevant anymore, removed. Resolving.

stijnmoreels:
Added a `MutateSecretNameCachedSecretProvider` too so we can combine caching and mutation together.

docs/preview/features/secret-store/provider/key-vault.md
```diff
@@ -48,6 +48,10 @@ public class Program
                         // Assing a configurable cached variant of the Azure Key Vault provider.
                         var cacheConfiguration = new CacheConfiguration(TimeSpan.FromMinutes(1));
                         builder.AddAzureKeyVaultWithManagedServiceIdentity(keyVaultUri, cacheConfiguration);
+
+                        // Adding the Azure Key Vault secret provider, using `.` instead of `:` when looking up secrets.
+                        // Example - When looking up `ServicePrincipal:ClientKey` it will be changed to `ServicePrincipal.ClientKey`.
```
tomkerkhove, line 53: [Using `.` is not allowed](https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#:~:text=Key%20Vault%20name%20must%20be,a%2Dz%2C%20A%2DZ%2C%20and%20%2D.&text=The%20type%20of%20the%20object,%22%2C%20or%20'certificates'.&text=An%20object%2Dname%20is%20a,unique%20within%20a%20Key%20Vault.) and one of the reasons I was asking for this 😎 

```suggestion
                        // Example - When looking up `ServicePrincipal:ClientKey` it will be changed to `ServicePrincipal-ClientKey`.
```

docs/preview/features/secret-store/provider/key-vault.md
```diff
@@ -48,6 +48,10 @@ public class Program
...
+                        builder.AddAzureKeyVaultWithManagedServiceIdentity(keyVaultUri, mutateSecretName: secretName => secretName.Replace(":", "."));
```
tomkerkhove, line 54:
```suggestion
                        builder.AddAzureKeyVaultWithManagedServiceIdentity(keyVaultUri, mutateSecretName: secretName => secretName.Replace(":", "-"));
```

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+        public async Task<string> GetRawSecretAsync(string secretName, bool ignoreCache)
+        {
+            Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Requires a non-blank secret name when mutating secret names");
+
+            string mutatedSecretName = MutateSecretName(secretName);
+            Task<string> rawSecretAsync = _implementation.GetRawSecretAsync(mutatedSecretName, ignoreCache);
+
+            if (rawSecretAsync is null)
+            {
+                throw new SecretNotFoundException(mutatedSecretName);
```
tomkerkhove, line 59: If this check is to ensure that a task is provided, is this exception then correct? Feels like it's not, no?

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            try
+            {
+                return await rawSecretAsync;
+            }
+            catch (Exception exception)
+            {
+                Logger.LogError(
```
tomkerkhove, line 68: Let's use a warning here and leave it up to the users to decide if this should trigger an alert or not given we will still throw an exception.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<Secret> GetSecretAsync(string secretName, bool ignoreCache)
+        {
+            Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Requires a non-blank secret name when mutating secret names");
+
+            string mutatedSecretName = MutateSecretName(secretName);
+            Task<Secret> secretAsync = _implementation.GetSecretAsync(mutatedSecretName, ignoreCache);
+
+            if (secretAsync is null)
+            {
```
tomkerkhove, line 92: Same here, correct exception or not?

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// Retrieves the secret value, based on the given name
+        /// </summary>
+        /// <param name="secretName">The name of the secret key</param>
+        /// <param name="ignoreCache">Indicates if the cache should be used or skipped</param>
+        /// <returns>Returns a <see cref="Task{TResult}"/> that contains the secret key</returns>
+        /// <exception cref="ArgumentException">The name must not be empty</exception>
+        /// <exception cref="ArgumentNullException">The name must not be null</exception>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<string> GetRawSecretAsync(string secretName, bool ignoreCache)
+        {
```
tomkerkhove, line 51: Shouldn't this one call `GetSecretAsync` to reduce duplication?

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+                throw new SecretNotFoundException(mutatedSecretName);
+            }
+
+            try
+            {
+                await invalidateSecretAsync;
+            }
+            catch (Exception exception)
+            {
+                Logger.LogError(
```
tomkerkhove, line 132: Same here, warning maybe?

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+                    exception, "Failure during invalidating secret '{MutatedSecretName}' that was mutated from '{OriginalSecretName}'", mutatedSecretName, secretName);
+
+                throw new SecretNotFoundException(mutatedSecretName, exception);
```
tomkerkhove, line 135: Is this correct? Who says it was because the secret was not found?

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+                throw new SecretNotFoundException(mutatedSecretName);
+            }
+
+            try
+            {
+                return await secretAsync;
+            }
+            catch (Exception exception)
+            {
+                Logger.LogError(
```
tomkerkhove, line 102: Same here, warning maybe?

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+using GuardNet;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Arcus.Security.Core.Providers
+{
+    /// <summary>
+    /// Represents an <see cref="ISecretProvider"/> that can mutate the secret name provided before looking up the secret.
+    /// </summary>
+    public class MutatedSecretNameSecretProvider : ISecretProvider
```
tomkerkhove, line 12: See feedback in MutatedSecretNameCachedSecretProvider. Maybe you can refactor shared things into MutatedSecretSecretProvider or so given this is duplicated here a bit as well.

tomkerkhove (approved):
LGTM, last round of feedback!

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Hmm, you may have a point, it may have to be `InvalidOperationException`? bc it's about the way the inner `ISecretProvider` is implemented? Good for you?

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
stijnmoreels: Yes, did that and reverted bc it may lead to 'harder' things to understand since we'll have to pass along functions as arguments to wrap it in an try/catch block. Not sure if it's more readable or less duplication that we're after.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Hmm, no bc we want to call the `GetRawSecretAsync` on the inner `ISecretProvider` otherwise the `GetRawSecretAsync` on the inner implementation will never be called.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
tomkerkhove: Sounds good!

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
tomkerkhove: Good point! Maybe we can change how we handle the task so that it's re-used and only call different operations on the secret provider here?

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
tomkerkhove: But the logic looks the mainly the same but it'sp probably the things it calls that are different?

For example, you can keep these and move how we handle the task so that it's re-used and only call different operations on the secret provider here? Just thinking out loud, feel free to resolve.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Moved to central method and changed the exception type. Resolving.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Changed to `LogWarning` and moved to central method.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Changed exception type and moved to central methd. Resolving.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Yes, exactly! Moved the common logic to a central method. Thanks. Resolving.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Changed to `LogWarning` and moved to central method. Resolving.

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
stijnmoreels: Yes, that's somewhat what I've done. It's now only different in the way the operations are called on the inner secret providers. Thanks! Resolving.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Changed to `LogWarning` and moved to central method. Resolving.

src/Arcus.Security.Core/Providers/MutatedSecretNameCachedSecretProvider.cs
stijnmoreels: Changed to `InvalidOperationException` exception type and moved to central method. Thanks! Resolving.

src/Arcus.Security.Core/Providers/MutatedSecretNameSecretProvider.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// <returns>
+        ///     The value that was returned from the <paramref name="asyncFuncAfterMutation"/> given the mutated version of the given <paramref name="secretName"/>.
+        /// </returns>
+        protected async Task<T> SafeguardMutateSecretNameAsync<T>(string secretName, Func<string, Task<T>> asyncFuncAfterMutation)
+        {
+            Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Requires a non-blank secret name when mutating secret names");
+            Guard.NotNull(asyncFuncAfterMutation, nameof(asyncFuncAfterMutation), "Requires a asynchronous function to execute after the secret name mutation");
+
+            string mutatedSecretName = MutateSecretName(secretName);
+            Task<T> secretAsync = asyncFuncAfterMutation(mutatedSecretName);
```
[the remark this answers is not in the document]
tomkerkhove: This is fine for me how it is.

Issue #149, opened 2020-06-23
Secret Provider for Docker secrets


fgheysels:
I think separating secrets from configuration is very valuable, and I was thinking if it would make sense to create a provider which can be used for Docker secrets.

**What secret store are you interested in?**
A secret store provider which allows to access docker secrets that are made available to a container.

**What scenarios would it enable you?**
It would enable me to treat docker secrets as secrets (and not as configuration values) in a docker container stack.

**More information**
There exists a Configuration Provider which makes docker secrets available in the configuration.
When giving a container access to a secret, the secret gets 'injected' as a file in the container.  The [KeyPerFileConfigurationProvider](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.keyperfile.keyperfileconfigurationprovider?view=dotnet-plat-ext-3.1) allows to make those files available in the configuration, however, since those things are actually secrets, it would make sense to create a Secret Provider for it.

Suppose you have a docker compose file which contains this:

```yaml
services:
  foo-service:
    image: localhost:5000/foo-service
    secrets:
      ConnectionStrings__FooDatabase

secrets:
  ConnectionStrings__FooDatabase:
    external: true
```

When running the container, a file will be injected in a certain location (the location depends whether you're targetting Linux or Windows containers) and that file will contain the secret.   It would be nice if there was a provider which is similar as the `KeyPerFileConfigurationProvider` which adds those files in the Secret Store.
```

tomkerkhove:
Fair ask, do you have a sample of a mounted secret file? Just to see if we should separate it from other secret providers which are file based (such as Key Vault + containers, HashiCorp Vault)

fgheysels:
Sure, they're just plain text files:

![image](https://user-images.githubusercontent.com/3605786/85384450-7645a980-b541-11ea-8b79-d9e95619ab11.png)

tomkerkhove:
Just curious - Are you using this outside of development as well?

fgheysels:
This is something I can pick up if you want; will do it in my spare time though

tomkerkhove:
That would be great, thanks! But if it's too much we can pull in @stijnmoreels as well.

Just let me know!

fgheysels:
I'll be glad to do it :)

fgheysels:
I've started implementing this.   First idea was to build this from scratch but then I saw that @stijnmoreels used the existing configuration package to implement another provider, so I've decided to go that route as well if that is ok for you.

[...]

PR #171, opened 2020-08-27
Add a SecretProvider that can be used for Docker Secrets 

fgheysels:
When using Docker Swarm & Docker secrets, secrets are mounted into the docker container as files.
This KeyPerFile secret provider (which is based on the [KeyPerFile](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.KeyPerFile/) configuration provider) allows easy access to those secrets.

(Maybe the secretprovider should be renamed to DockerSecrets ?  I've choosen KeyPerFile since it is based on that package ... )

This is related to #149


tomkerkhove:
@stijnmoreels Do you want to pick this review up?

Looks good from a first glance, only thing I'm not sure about is the naming indeed. I'm more on the DockerSecrets front as it's more straight forward for what it's built as per [the docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#key-per-file-configuration-provider) but can also see why you'd not want to do it 🤔

stijnmoreels:
> @stijnmoreels Do you want to pick this review up?

Yes, of course, will do it first thing monday, though if that's OK. I'll assign myself.

docs/features/secret-store/provider/key-per-file.md
```diff
@@ -0,0 +1,88 @@
+---
```
stijnmoreels, line 1: Since this is not currentlyl released yet, this feature docs should be moved to `/docs/preview/features/secret-store/provider/key-per-file.md`.

docs/features/secret-store/provider/key-per-file.md
```diff
@@ -0,0 +1,88 @@
...
+title: "KeyPerFile secret provider"
+layout: default
+---
+
+# KeyPerFile secret provider
+The _KeyPerFile_  secret provider allows working with Docker secrets.  When using Docker secrets in Docker Swarm, the secrets are injected in the Docker container as files.  
+The _KeyPerFile_ secret provider provides access to those secrets via the `ISecretStore`.
```
stijnmoreels, line 8:
```suggestion
The _KeyPerFile_ secret provider provides access to those secrets via the secret store.
```

We don't have a `ISecretStore` interface. Maybe just call it 'secret store'.

docs/features/secret-store/provider/key-per-file.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+After adding the _KeyPerFile_ secret provider to the secret store, the Docker secrets can simply be retrieved by calling the appropriate methods on the `ISecretProvider`:
+
+```csharp
+public class PersonController
+{
+    private readonly ISecretProvider _secrets;
+
+    public PersonController( ISecretProvider secrets )
```
stijnmoreels, line 69:
```suggestion
    public PersonController(ISecretProvider secrets )
```

Nitpic, unnecessary space 😄 .

docs/features/secret-store/provider/key-per-file.md
```diff
@@ -0,0 +1,88 @@
...
+    {
+        _secrets = secrets;
+    }
+
+    [HttpGet]
+    public async Task GetPerson( Guid personId )
```
stijnmoreels, line 75:
```suggestion
    public async Task GetPerson(Guid personId)
```

Nitpick, unnecessary space 😄 .

docs/features/secret-store/provider/key-per-file.md
```diff
@@ -0,0 +1,88 @@
...
+    {
+        var connectionstring = _secrets.GetRawSecretAsync("ConnectionStrings:PersonDatabase")
+
+        using( var connection = new SqlDbConnection(connectionstring))
```
stijnmoreels, line 79:
```suggestion
        using (var connection = new SqlDbConnection(connectionstring))
```

Nitpick, unnecessary space 😄 . Following some code style Arcus guidelines and consistency.

src/Arcus.Security.Providers.KeyPerFile/Arcus.Security.Providers.KeyPerFile.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.1</TargetFrameworks>
+  </PropertyGroup>
```
stijnmoreels, line 5:
```suggestion
<PropertyGroup>
    <TargetFrameworks>netstandard2.0;netcoreapp3.1</TargetFrameworks>
    <Authors>Arcus</Authors>
    <Description>Provides support for Docker Secrets</Description>
    <Copyright>Copyright (c) Arcus</Copyright>
    <PackageLicenseUrl>https://github.com/arcus-azure/arcus.security/blob/master/LICENSE</PackageLicenseUrl>
    <PackageProjectUrl>https://github.com/arcus-azure/arcus.security</PackageProjectUrl>
    <PackageIconUrl>https://raw.githubusercontent.com/arcus-azure/arcus/master/media/arcus.png</PackageIconUrl>
    <RepositoryUrl>https://github.com/arcus-azure/arcus.security</RepositoryUrl>
    <RepositoryType>Git</RepositoryType>
    <PackageTags>Docker;User Secrets Manager;Security;Key Per FIle</PackageTags>
    <AssemblyName>Arcus.Security.Providers.KeyPerFile</AssemblyName>
    <RootNamespace>Arcus.Security.Providers.KeyPerFile</RootNamespace>
    <PackageId>Arcus.Security.Providers.KeyPerFile</PackageId>
    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
    <GenerateDocumentationFile>true</GenerateDocumentationFile>
  </PropertyGroup>
```

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
```diff
@@ -0,0 +1,60 @@
+﻿using Arcus.Security.Core;
+using GuardNet;
+using Microsoft.Extensions.Configuration.KeyPerFile;
+using System;
+using System.Threading.Tasks;
+
+namespace Arcus.Security.Providers.KeyPerFile
+{
+    public class KeyPerFileSecretProvider : ISecretProvider
```
stijnmoreels, line 9:
```suggestion
    /// <summary>
    /// Represents an <see cref="ISecretProvider" /> that provides access to the Docker secrets mounted into the Docker container as files.
    /// </summary>
    public class KeyPerFileSecretProvider : ISecretProvider
```

Provide some summary for the secret provider.

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        private readonly KeyPerFileConfigurationProvider _provider;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="KeyPerFileSecretProvider"/> class.
+        /// </summary>
```
stijnmoreels, line 15:
```suggestion
        /// <summary>
        /// Initializes a new instance of the <see cref="KeyPerFileSecretProvider"/> class.
        /// </summary>
        /// <param name="configurationSource">The configuration source that provides the Docker secrets mounted as files in the container.</param>
        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="configurationSource"/> is <c>null</c></exception>
```

Provides some argument information.

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        public KeyPerFileSecretProvider(KeyPerFileConfigurationSource configurationSource)
+        {
+            Guard.NotNull(configurationSource, nameof(configurationSource));
+            _provider = new KeyPerFileConfigurationProvider(configurationSource);
+            _provider.Load();
```
stijnmoreels, line 20: Reason why this is called inside the constructor and not in the extension? Because it allows you to make sure that the configuration provider was loaded?

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// Retrieves the secret value, based on the given name
+        /// </summary>
+        /// <param name="secretName">The name of the secret key</param>
+        /// <returns>Returns the secret key.</returns>
+        /// <exception cref="ArgumentException">The <paramref name="secretName"/> must not be empty</exception>
+        /// <exception cref="ArgumentNullException">The <paramref name="secretName"/> must not be null</exception>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public Task<string> GetRawSecretAsync(string secretName)
+        {
+            if (_provider.TryGet(secretName, out string value))
```
stijnmoreels, line 33:
```suggestion
        public Task<string> GetRawSecretAsync(string secretName)
        {
            Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Requires a non-blank secret name to retrieve Docker secrets");
            
            if (_provider.TryGet(secretName, out string value))
```

Let's provider a guard here.

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// Retrieves the secret value, based on the given name
+        /// </summary>
+        /// <param name="secretName">The name of the secret key</param>
+        /// <returns>Returns a <see cref="Secret"/> that contains the secret key</returns>
+        /// <exception cref="ArgumentException">The <paramref name="secretName"/> must not be empty</exception>
+        /// <exception cref="ArgumentNullException">The <paramref name="secretName"/> must not be null</exception>
+        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        public async Task<Secret> GetSecretAsync(string secretName)
+        {
+            string secretValue = await GetRawSecretAsync(secretName);
```
stijnmoreels, line 51:
```suggestion
        public async Task<Secret> GetSecretAsync(string secretName)
        {
            Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Requirs a non-blank secret name to retrieve Docker secrets");
            
            string secretValue = await GetRawSecretAsync(secretName);
```

Let's guard here.

docs/features/secret-store/provider/key-per-file.md
fgheysels: Yeah, that's how my coding style in VS.NET is configured.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.Extensions.Configuration.KeyPerFile;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Hosting;
+
+namespace Arcus.Security.Providers.KeyPerFile
```
stijnmoreels, line 5:
```suggestion
// ReSharper disable once CheckNamespace
namespace Microsoft.Extensions.Hosting
```

We tend to use the namespace of the targetted type as the namespace for the extension.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public static class SecretStoreBuilderExtensions
```
stijnmoreels, line 7:
```suggestion
    /// <summary>
    /// Extensions on the <see cref="SecretStoreBuilder" /> related to Docker secrets.
    /// </summary>
    public static class SecretStoreBuilderExtensions
```

Public types documented, maybe.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public static SecretStoreBuilder AddKeyPerFile(this SecretStoreBuilder builder, string directoryPath, bool optional = true)
```
stijnmoreels, line 9:
```suggestion
        /// <summary>
        /// Adds Docker secrets mounted as files in the Docker container to the secret store.
        /// </summary>
        /// <param name="builder">The builder to add the Docker secrets provider to.</param>
        /// <param name="directoryPath">The direcotry path where the Docker secrets are located.</param>
        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c></exception>
        /// <exception cref="ArgumentException">Throw when the <paramref name="directoryPath"/> is blank</exception>
        public static SecretStoreBuilder AddKeyPerFile(this SecretStoreBuilder builder, string directoryPath, bool optional = true)
```

Some public docs.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+        {
+            KeyPerFileConfigurationSource configuration = new KeyPerFileConfigurationSource();
```
stijnmoreels, line 11:
```suggestion
        {
            Guard.NotNull(builder, nameof(builder), "Requires a secret store builder to add the Docker secrets to");
            Guard.NotNullOrWhitespace(directoryPath, nameof(directoryPath), "Requires a non-blank directory path to locate the Docker secrets");
            
            KeyPerFileConfigurationSource configuration = new KeyPerFileConfigurationSource();
```

Let's guard here.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
stijnmoreels, line 9: Should we use the `optional` here? It was supposed to be used (I think) to return `null` instead of throwing, but because we use the secret store, this feature is resulting in the same result bc the secret store will eventuall throw if the secret was not found in any of the registered providers.

src/Arcus.Security.Tests.Integration/KeyPerFile/SecretStoreBuilderExtensionTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.Extensions.Hosting;
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Arcus.Security.Tests.Integration.KeyPerFile
+{
+    public class SecretStoreBuilderExtensionTests : IntegrationTest, IDisposable
```
stijnmoreels, line 13: Two positive tests, maybe some negatives are useful here too. Thanks.

src/Arcus.Security.Tests.Integration/KeyPerFile/SecretStoreBuilderExtensionTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            IHost host = hostBuilder.Build();
+            var secretProvider = host.Services.GetRequiredService<ISecretProvider>();
+
+            string actualValue = await secretProvider.GetRawSecretAsync("ConnectionStrings:PersonDb");
+            Assert.Equal(expectedValue, actualValue);
+        }
+
+        private async Task SetSecretAsync(string secretKey, string secretValue)
+        {
+            await File.WriteAllTextAsync(Path.Combine(_secretLocation, secretKey), secretValue);
```
stijnmoreels, line 66:
```suggestion
            string filePath = Path.Combine(_secretLocation, secretKey);
            await File.WriteAllTextAsync(filePath, secretValue);
```

Apparently, we should use in Arcus less inline stuff 😄 , sorry  about that.

stijnmoreels (changes requested):

stijnmoreels:
Thanks @fgheysels for the implementation, clearly specified docs and tests! Very clear, and it shows that you try to follow the Arcus standards. Thanks for this!

docs/features/secret-store/provider/key-per-file.md
stijnmoreels: Togheter with moving it to the `/docs/preview/` path, you may want to add a link to the `/docs/preview/index.md` that pionts to this file.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
stijnmoreels, line 9: I'll have to check with @tomkerkhove , but I think we can benefit from having also here a secret name mutation optional parameter.  See [this link](https://github.com/arcus-azure/arcus.security/blob/03a7d85b4f3b44ee1a1c2c7757202167e8357606/src/Arcus.Security.Providers.UserSecrets/Extensions/SecretStoreBuilderExtensions.cs#L30) on how this is done for the User Secrets. It makes it easier maintenance for devs. (Plus, some tests are in order when this optional parameter is added of course).

docs/features/secret-store/provider/key-per-file.md
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+```
+
+## Retrieving secrets
+
+Suppose you have the following docker-compose file:
+
+```yaml
+version: '3.8'
```
stijnmoreels, line 48: Any change we can set up a Docker project and make an integration test for this? (Maybe diff PR if too much work).

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
fgheysels: It makes sure that if I instantiate an instance, the secrets are loaded as well.
It might make sense to move it outside the ctor, but then users will to call the Load method themselves.   I can change that if you want

tomkerkhove: This is ok for me! But does this mean it will not reflect changes to the file later on?

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
tomkerkhove: Yes indeed. If I use "foo-bar" as a pattern, it would be good to be able to do "foo-bar" --> "foo:bar"

docs/features/secret-store/provider/key-per-file.md
tomkerkhove: Let's do seperate PR indeed so you can have a look at it and we don't put the burden on @fgheysels indeed, if ok?

fgheysels:
> Thanks @fgheysels for the implementation, clearly specified docs and tests! Very clear, and it shows that you try to follow the Arcus standards. Thanks for this!

Seems like my VS.NET settings / preferences regarding code layout are not conform with the Arcus guidelines :)

fgheysels:
> I'm more on the DockerSecrets front as it's more straight forward for what it's built as per [the docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#key-per-file-configuration-provider) but can also see why you'd not want to do it 🤔

Shall I rename it ?

tomkerkhove:
It's up to you but it would feel more user-friendly to me but not mandatory at all!

fgheysels:
> It's up to you but it would feel more user-friendly to me but not mandatory at all!

I'll rename it.
I'll do it later this week, and will also have a look at resolving the remarks made by @stijnmoreels  later this week

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
stijnmoreels: The UserSecrets is different in that way. It's begin called there before the ctor. Seems like they both should have the same setup but not sure which one.
Not sure also if it should be the responsibility of the secret provider that the configuration was loaded. 🤔 If this could be checked or so... Also, with 'reloading' this should maybe looked further into.

fgheysels: Well, the ´Load´ method is part of the KeyPerFileConfigurationProvider whereas the parameter of the ctor is a configuration source.
I could change that we're injecting a configurationprovider instead of a configuration source though.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
fgheysels: If optional is set to false, an exception will be thrown when the directory-path (where the secrets are located inside the container) doesn't exist.
If set to true, no exception will be thrown at that point, however, an exception will be thrown later (which says that the secret could not be found).   However, if we have multiple secretstores configured, the secret might as well be found by another secret provider ....
Don't know what the best approach is here, therefore I've just taken over the functionality that exists in the 'KeyPerFile' config provider.

docs/features/secret-store/provider/key-per-file.md
fgheysels: How do you see that implemented ?

src/Arcus.Security.Providers.KeyPerFile/Arcus.Security.Providers.KeyPerFile.csproj
fgheysels: Done.  I've chosen some different tasks though instead of the ones you've suggested.

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
fgheysels: Done.   I did it via VS.NET instead of committing the suggestion, since I've renamed some files and directories.

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
fgheysels: Done.   I did it via VS.NET instead of committing the suggestion, since I've renamed some files and directories.

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
fgheysels: Done, but with a slightly other message.

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
fgheysels: Done, but the message is slightly different

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
fgheysels: Makes sense.  Done.  (Couldn't do this via the suggestion since the files have changed + also had impact on the used namespaces).

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
fgheysels: Done.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
fgheysels: Changed the name of the method as well.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
fgheysels: Done.   I did it via VS.NET instead of committing the suggestion, since I've renamed some files and directories.

fgheysels:
> It's up to you but it would feel more user-friendly to me but not mandatory at all!

Renamed!

docs/features/secret-store/provider/key-per-file.md
fgheysels: Done.  I've also updated the contents of the documentation file.

fgheysels:
> > It's up to you but it would feel more user-friendly to me but not mandatory at all!
> 
> I'll rename it.
> I'll do it later this week, and will also have a look at resolving the remarks made by @stijnmoreels later this week

Seems like a week became a month. :)

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
fgheysels: The underlying Config Provider already does a translation from "foo:bar" to "foo__bar", see [this test](https://github.com/fgheysels/arcus.security/blob/frgh/feature/149_dockersecrets/src/Arcus.Security.Tests.Integration/DockerSecrets/SecretStoreBuilderExtensionTests.cs#L42), as docker secrets use the double underscore as separator.

Or are you talking about something else ?

tomkerkhove: We provide overloads where you can provide your own translation. Maybe they are using "foo-bar" in the app and you want to look it up in the Docker Secrets through "foo:bar".

tomkerkhove: If it's just for the file then I would use false internally because otherwise there is a "setup issue", but I might be too extreme?

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
tomkerkhove: Would we then have to still load it or just act as we call it?

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
stijnmoreels: Yes, I agree with @tomkerkhove , our other providers are also throwing on 'setup' things and if we would want to have such an optional functionality, I think we should look into it on a broader level and how we can use it for every provider. So, for now, maybe set it on `false` and we can maybe look at it in an issue later.

docs/features/secret-store/provider/key-per-file.md
stijnmoreels: Something like starting the Docker project with ebmedded secrets beforehand, or don't we need a Docker project to test this? Can we just write directly to a file like the UserSecrets? Than it would be even more easy.

docs/preview/index.md
```diff
@@ -24,6 +24,7 @@ PM > Install-Package Arcus.Security.Providers.AzureKeyVault
     - [Azure Key Vault](features/secret-store/provider/key-vault)
     - [Configuration](features/secret-store/provider/configuration)
     - [Environment variables](features/secret-store/provider/environment-variables)
+    - [Docker secrets](features/secret-store/provider/docker-secrets.md)
```
stijnmoreels, line 27:
```suggestion
    - [Docker secrets](features/secret-store/provider/docker-secrets)
```

Links don't need an `.md` extension.

src/Arcus.Security.Providers.DockerSecrets/DockerSecretsSecretProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DockerSecretsSecretProvider"/> class.
+        /// </summary>
+        /// <param name="configurationSource">The configuration source that provides the Docker secrets mounted as files in the container.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="configurationSource"/> is <c>null</c></exception>
+        public DockerSecretsSecretProvider(KeyPerFileConfigurationSource configurationSource)
+        {
+            Guard.NotNull(configurationSource, nameof(configurationSource));
+            _provider = new KeyPerFileConfigurationProvider(configurationSource);
+            _provider.Load();
```
stijnmoreels, line 25: So, it's OK to do it here, but we should make sure that it's the same approach as the UserSecrets. So either change it here or change it in the UserSecrets. That would be best.

stijnmoreels (changes requested):

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
fgheysels: I've changed and pushed this, but would like to do some additional testing.

tomkerkhove: No problem, let us know when you are ready!

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
fgheysels: What do you mean ?  Call load when the first secret is retrieved ?

tomkerkhove: Yes or will it work automatically with ConfigurationProvider? Or does that depend on the IConfiguration of the user? Sorry I'm lost here I think.

fgheysels: Calling `Load` on the provider loads all files that are present in the configured directory, and creates 'secrets' from them.  So, if you create the secret provider, all secrets will be loaded only once.  
In contrast of the `KeyVaultSecretProvider`which requests the secret every time from KeyVault itself (when no caching is configured), this provider retrieves the secrets only once.

This can be changed offcourse, but  then we should not rely on the `KeyPerFile` configuration provider which is now used internally, but implement this functionality ourselves.  In that case, we would need to load the correct file, read and return its contents on every call to 'GetSecretAsync'.

src/Arcus.Security.Providers.KeyPerFile/SecretStoreBuilderExtensions.cs
fgheysels: I've changed this, but now this means that if you specify a non-existing directory, the container will fail to start.  That is a good example of failing fast but I don't know if you really want that.

What if you have a program which has multiple secret providers configured.   In some cases, your program might not be running in a container and needs to retrieve secrets from another location.  In that case, you still need to have that directory present.

src/Arcus.Security.Providers.DockerSecrets/DockerSecretsSecretProvider.cs
fgheysels: I've changed this even more.
In my opinion, the only thing that users of the `DockerSecretProvider` should be aware of, is the location where the secrets are stored inside the container.  How those secrets are being retrieved, is an implementation detail and users of `DockerSecretProvider`shouldn't be bothered with that.

Therefore, I've changed the constructor so that we only need to pass in the directory where the secrets are located.  The fact that we're now using the `KeyPerFileProvider` is an implementation detail and should be hidden from the callers.   If we would change the implementation of the `DockerSecretProvider` (f.i. no longer rely on the `KeyPerFileConfigProvider`), that should be transparent for the users of the `DockerSecretProvider`, and they should not be confronted with having to perform code changes for that.

src/Arcus.Security.Providers.DockerSecrets/SecretStoreBuilderExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <summary>
+        /// Adds Docker secrets (mounted as files in the Docker container) to the secret store.
+        /// </summary>
+        /// <param name="builder">The builder to add the Docker secrets provider to.</param>
+        /// <param name="directoryPath">The path inside the container where the Docker secrets are located.</param>
+        /// <param name="mutateSecretName">The optional function to mutate the secret name before looking it up.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c></exception>
+        /// <exception cref="ArgumentException">Throw when the <paramref name="directoryPath"/> is blank</exception>
+        public static SecretStoreBuilder AddDockerSecrets(this SecretStoreBuilder builder, string directoryPath, Func<string, string> mutateSecretName = null)
```
fgheysels, line 23: I've added a `mutateSecretName` argument, but I'm wondering if we should provide a default implementation here, instead of having 'null' as the default value ?

Default implementation could be `secretName => secretName.Replace(":", "__")` for instance.

src/Arcus.Security.Tests.Integration/KeyPerFile/SecretStoreBuilderExtensionTests.cs
fgheysels: Added some additional tests

src/Arcus.Security.Providers.DockerSecrets/DockerSecretsSecretProvider.cs
stijnmoreels: Ah, yes, I can agree on that. Can you create an issue for this to change it in the UserSecrets provider? Thanks. I'll resolve this.

docs/preview/features/secret-store/provider/docker-secrets.md
```diff
@@ -0,0 +1,90 @@
...
+
+    public static IHostBuilder CreateHostBuilder(string[] args)
+    {    
+        return Host.CreateDefaultBuilder(args)
+                   .ConfigureSecretStore((context, config, builder) =>
+                    {
+                        // Adds the secrets that exist in the "/run/secrets" directory to the ISecretStore
+                        // Docker secrets are by default mounted into the /run/secrets directory
+                        // when using Linux containers on Docker Swarm.
+                        builder.AddDockerSecrets(directoryPath: "/run/secrets", optional: true);
```
stijnmoreels, line 38:
```suggestion
                        builder.AddDockerSecrets(directoryPath: "/run/secrets");
```

Is removed.

src/Arcus.Security.Providers.DockerSecrets/DockerSecretsSecretProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DockerSecretsSecretProvider"/> class.
+        /// </summary>
+        /// <param name="secretsDirectoryPath">The path inside the docker container where the secrets are located.</param>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="secretsDirectoryPath"/> is <c>null</c>or an empty string.</exception>
+        public DockerSecretsSecretProvider(string secretsDirectoryPath)
+        {
+            Guard.NotNullOrWhitespace(secretsDirectoryPath, nameof(secretsDirectoryPath));
+
```
stijnmoreels, line 25: Should we here check if the reference to the secret directory path points to a existing directory? Or should we let the remaining code do that. I'm only thinking that the exception thrown there will maybe more obscure and will not include the information that we're in the `DockerSecretProvider`.

src/Arcus.Security.Providers.DockerSecrets/DockerSecretsSecretProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    public class DockerSecretsSecretProvider : ISecretProvider
+    {
+        private readonly KeyPerFileConfigurationProvider _provider;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DockerSecretsSecretProvider"/> class.
+        /// </summary>
+        /// <param name="secretsDirectoryPath">The path inside the docker container where the secrets are located.</param>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="secretsDirectoryPath"/> is <c>null</c>or an empty string.</exception>
```
stijnmoreels, line 21:
```suggestion
        /// <exception cref="ArgumentException">Thrown when the <paramref name="secretsDirectoryPath"/> is blank.</exception>
```

We also throw on blank `string`s 😉 .

src/Arcus.Security.Tests.Integration/DockerSecrets/SecretStoreBuilderExtensionTests.cs
```diff
@@ -0,0 +1,85 @@
...
+            // Assert
+            IHost host = hostBuilder.Build();
+            var secretProvider = host.Services.GetRequiredService<ISecretProvider>();
+
+            string actualValue = await secretProvider.GetRawSecretAsync(secretKey);
+            Assert.Equal(expectedValue, actualValue);
+        }
+
+        [Fact]
+        public async Task DockerSecretsProvider_ReturnsNull_WhenSecretNotFound()
```
stijnmoreels, line 44: I'm sorry for the thorough comments. I was thinking here. Maybe also a test where we set a secret to a docker secret file but request a secret with a name that isn't present in that secret file, so we can check that the selective searching is correct.

stijnmoreels (changes requested):
We're getting there. Will be an amazing addition to our secret store. Thanks!

src/Arcus.Security.Providers.DockerSecrets/SecretStoreBuilderExtensions.cs
stijnmoreels: Hmm, we could discus this maybe in an issue. You could argue that you may want to set some general secret name mutation on the secret store -level instead on each provider. 🤔

docs/preview/features/secret-store/provider/docker-secrets.md
fgheysels: That has been overlooked, thx!.

src/Arcus.Security.Providers.DockerSecrets/SecretStoreBuilderExtensions.cs
fgheysels: I wouldn't opt for a general mutation on secret store level, as I feel that the mutation should be specific for providers.

When you have a look at how configuration values are retrieved from the Configuration, your code typically uses `:` as the 'hierarchy separator'.  When you want to retrieve secrets, and you have multiple secret providers configured, your calling code also does not need to know which 'mutation' it needs to do.   That's why we specify this on provider level.
For instance, if I have configured the `KeyVaultProvider` and the `DockerSecretsProvider` , I want to be able to retrieve a secret by just calling `GetSecret("ConnectionStrings:MyDatabase")`.   When the `KeyVaultProvider` needs to retrieve the secret from KeyVault, the secretname needs to be changed to `ConnectionStrings--MyDatabase`.   When the secret is retrieved from the Docker secrets, the name must be changed to `ConnectionStrings__MyDatabase`.

(Maybe this translation should not even be defined via the `mutateSecret` func, but should be baked into the implementation of the `ISecretProvider` ?)

src/Arcus.Security.Tests.Integration/DockerSecrets/SecretStoreBuilderExtensionTests.cs
fgheysels: I can do that, somewhere later this week probably.

src/Arcus.Security.Providers.DockerSecrets/SecretStoreBuilderExtensions.cs
fgheysels: But you're right.  We can discuss in a separate issue.  I'll open one later today.

src/Arcus.Security.Providers.DockerSecrets/DockerSecretsSecretProvider.cs
fgheysels: True.  If added a check and throw an exception in the DockerSecretsProvider if the directory doesn't exist.

However, I'm still not convinced if throwing an exception (and using Optional=false) is the approach that we need to take.  See also this comment: https://github.com/arcus-azure/arcus.security/pull/171#discussion_r502911108

src/Arcus.Security.Providers.DockerSecrets/SecretStoreBuilderExtensions.cs
fgheysels: Started a new issue for this: https://github.com/arcus-azure/arcus.security/issues/199

src/Arcus.Security.Tests.Integration/DockerSecrets/SecretStoreBuilderExtensionTests.cs
fgheysels: I've changed the existing testcase.   In that test, I just make sure that a secret exists.

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
fgheysels: @tomkerkhove what is your view on this ?

src/Arcus.Security.Tests.Integration/DockerSecrets/SecretStoreBuilderExtensionTests.cs
stijnmoreels: Yeah, but maybe both are intressting. Right? Checking that the secret provider works both for an existing Docker secrets file and without.

src/Arcus.Security.Providers.DockerSecrets/DockerSecretsSecretProvider.cs
stijnmoreels: Yes, lets talk about that in here #200.

docs/preview/features/secret-store/provider/docker-secrets.md
```diff
@@ -0,0 +1,90 @@
...
+
+    public PersonController(ISecretProvider secrets)
+    {
+        _secrets = secrets;
+    }
+
+    [HttpGet]
+    public async Task GetPerson(Guid personId)
+    {
+        var connectionstring = _secrets.GetRawSecretAsync("ConnectionStrings:PersonDatabase")
```
stijnmoreels, line 79:
```suggestion
        string connectionstring = await _secrets.GetRawSecretAsync("ConnectionStrings:PersonDatabase")
```

😄

stijnmoreels (approved):
When my last comment is resolved, I think this is OK. Thanks for keeping up with this and the complete and correct addition! Much appriciated!

src/Arcus.Security.Tests.Integration/DockerSecrets/SecretStoreBuilderExtensionTests.cs
fgheysels: I don't see much difference ?  Each secret has it's own file.  So if I want to retrieve a secret that doesn't exist, the file will not be there.
I don't see the added value of having 2 similar tests:  one where we have no secrets (and thus no secretfiles) and a test where we do have some secret files, but want to retrieve a secret that doesn't exist

docs/preview/features/secret-store/provider/docker-secrets.md
fgheysels: We should have a compiler for docs.

src/Arcus.Security.Tests.Integration/DockerSecrets/SecretStoreBuilderExtensionTests.cs
stijnmoreels: Yeah, I think it's valuable bc you can prove that with or without the file it results in the same outcome. And it doesn't retrieve another file or the wrong one. And it doesn't affect the way the provider works. Maybe I would write something that would write a number of random secrets and retrieve a subset of them , but that would be a property test and not an integration test so this is fine.
But, maybe it's not necessary. I'll resolve this.

fgheysels:
I see there are some failing tests; I can have a look at it tonight

fgheysels:
> I see there are some failing tests; I can have a look at it tonight

Should be fixed now.

fgheysels:
Are there any issues left, or are we good to go ?

stijnmoreels:
> Are there any issues left, or are we good to go ?

I think not (as for my part, I guess). Let's see if @tomkerkhove has any remaining remarks.

docs/preview/index.md
```diff
@@ -24,6 +24,7 @@ PM > Install-Package Arcus.Security.Providers.AzureKeyVault
     - [Azure Key Vault](features/secret-store/provider/key-vault)
     - [Configuration](features/secret-store/provider/configuration)
     - [Environment variables](features/secret-store/provider/environment-variables)
+    - [Docker secrets](features/secret-store/provider/docker-secrets)
```
tomkerkhove, line 27: Nit - Can you move it before Environment variables please?

docs/preview/features/secret-store/provider/docker-secrets.md
```diff
@@ -0,0 +1,90 @@
+---
+title: "Docker Secrets secret provider"
+layout: default
+---
+
+# Docker Secrets secret provider
+The _DockerSecrets_  secret provider allows working with Docker secrets.  When using Docker secrets in Docker Swarm, the secrets are injected in the Docker container as files.
```
tomkerkhove, line 7: Feels more natural to read maybe?
```suggestion
This provider allows you to work with Docker secrets. When using Docker secrets in Docker Swarm, the secrets are injected in the Docker container as files.  
```

docs/preview/features/secret-store/provider/docker-secrets.md
```diff
@@ -0,0 +1,90 @@
+---
+title: "Docker Secrets secret provider"
+layout: default
+---
+
+# Docker Secrets secret provider
+The _DockerSecrets_  secret provider allows working with Docker secrets.  When using Docker secrets in Docker Swarm, the secrets are injected in the Docker container as files.  
+The _DockerSecrets_ secret provider provides access to those secrets via the secret store.
```
tomkerkhove, line 8:
```suggestion
The Docker secrets secret provider provides access to those secrets via the secret store.
```

docs/preview/features/secret-store/provider/docker-secrets.md
```diff
@@ -0,0 +1,90 @@
...
+        - 5555:80
+    secrets:
+        - ConnectionStrings__PersonDatabase
+
+secrets:
+  ConnectionStrings__PersonDatabase:
+    external: true
+```
+
+After adding the _DockerSecrets_ secret provider to the secret store, the Docker secrets can simply be retrieved by calling the appropriate methods on the `ISecretProvider`:
```
tomkerkhove, line 64:
```suggestion
After adding the Docker Secrets secret provider to the secret store, the Docker secrets can simply be retrieved by calling the appropriate methods on the `ISecretProvider`:
```

tomkerkhove (approved):
LGTM, just a few remarks for the docs

src/Arcus.Security.Providers.KeyPerFile/KeyPerFileSecretProvider.cs
tomkerkhove: > Question is, is this necessary ? When using docker and docker secrets, the secrets are mounted as files inside the container. When the secret changes, I guess that this also means that the container needs to be restarted ?

That is not the case actually, often hot-swapping of configs are done without restarting the container.

Maybe odd question, but what if we call Load now and then to refresh things?

docs/preview/index.md
fgheysels: Ah, didn't know they had to be in alphabetical order.  Changed.

tomkerkhove (approved):

stijnmoreels (approved):

PR #188, opened 2020-09-18
Feature - add secret name mutation to HashiCorp Vault secret store builder extensions

stijnmoreels:
Add secret name mutation to all the HashiCorp Vault secret store builder so the secret name can be altered before it's send to the HashiCorp Vault.

### DIFF
--- a/src/Arcus.Security.Tests.Integration/HashiCorp/HashiCorpSecretProviderTests.cs
+++ b/src/Arcus.Security.Tests.Integration/HashiCorp/HashiCorpSecretProviderTests.cs
@@ -1,9 +1,13 @@
 ﻿using System.Collections.Generic;
 using System.Threading.Tasks;
+using Arcus.Security.Core;
 using Arcus.Security.Providers.HashiCorp;
+using Arcus.Security.Providers.HashiCorp.Extensions;
 using Arcus.Security.Tests.Integration.Fixture;
 using Arcus.Security.Tests.Integration.HashiCorp.Hosting;
 using Arcus.Testing.Logging;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Vault.Endpoints.Sys;
 using VaultSharp;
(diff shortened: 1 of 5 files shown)


src/Arcus.Security.Tests.Integration/HashiCorp/HashiCorpSecretProviderTests.cs
```diff
@@ -65,6 +64,172 @@ public async Task AuthenticateWithUserPassKeyValueV2_GetSecret_Succeeds()
...
+                    mountPoint: DefaultDevMountPoint,
+                    path: secretPath,
+                    data: new Dictionary<string, object> { [secretKey] = expected });
+
+                // Act
+                builder.ConfigureSecretStore((config, stores) =>
+                {
+                    stores.AddHashiCorpVaultWithUserPass(
+                        server.ListenAddress.ToString(), userName, password, secretPath, keyValueMountPoint: DefaultDevMountPoint, 
+                        mutateSecretName: secretName => secretName.Remove(0, 5));
```
tomkerkhove, line 119: Maybe use replace instead so that we can use a variable which is the same as [here](https://github.com/arcus-azure/arcus.security/pull/188/files#diff-7a06e78fdfe87689905c17dc86d12041R125)?
```suggestion
                        mutateSecretName: secretName => secretName.Replace(SecretNamePrefix, string.Empty));
```

tomkerkhove (approved):

src/Arcus.Security.Tests.Integration/HashiCorp/HashiCorpSecretProviderTests.cs
stijnmoreels: Hmm, well, it's just by accident that they share that value. Both test (should not) have anything in common as to say that they only can be used like that. I think it will make it less clear.

tomkerkhove: But they are part of same test or what am I missing? 🤔 

We assume removing 0 to 5 is enough, but if we change "Test-" to something else it will break the test

stijnmoreels: Yes, that's why it should stay local. Maybe I'll change the test so they both mutate something different to avoid the confustion.

stijnmoreels: Ah, you mean local, you mean just a variable 🙄 . Ok, I can do that. 😄 .

tomkerkhove: Yes, sorry!

Issue #189, opened 2020-09-22
Arcus.Security has a dependency on Microsoft.AspNetCore.App


fgheysels:
**Describe the bug**
The `Arcus.Security.Core` and `Arcus.Security.Providers.AzureKeyVault` projects both have a dependency on `Microsoft.AspNetCore.App` which is in my humble opinion, and as far as I know, not necessary.    (I presume other projects inside the solution have the same dependency, but I haven't checked them all).

This creates an unnecessary dependency which is not a problem or does not become apparent as long as you use the `Arcus.Security` library in ASP.NET core projects.
I've created a console application however which runs as a container and consumes a message bus.  This project uses `Arcus.Security` to retrieve secrets using `ISecretProvider`.   As this container doesn't require any ASP.NET core features, the container is based on the `mcr.microsoft.com/dotnet/core/runtime:3.1-alpine`, but since `Arcus.Security` has a dependency on ASP.NET Core, my container fails to start:

```
It was not possible to find any compatible framework version
The framework 'Microsoft.AspNetCore.App', version '3.1.0' was not found.
  - No frameworks were found.
```

I think it might be sufficient to just remove the dependency on AspNetCore in the csproj files:

```csproj
<ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
    <FrameworkReference Include="Microsoft.AspNetCore.App" />
</ItemGroup>
```` 

I think the above itemgroups can be removed.

stijnmoreels:
Hmm, I thought that it was a previous concern that the double target framework on these projects would require us to use that framework reference. We have project templates that works fine without the ASP.NET Core reference [here](https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.ServiceBus.Topic/Dockerfile). I should ask @tomkerkhove for more info on this 🤔 .

tomkerkhove:
This is only required for .NET Core >3.x so we can remove it here.

The NuGet versioning scheme implies that as well: https://www.nuget.org/packages/Microsoft.AspNetCore.App/

fgheysels:
> Hmm, I thought that it was a previous concern that the double target framework on these projects would require us to use that framework reference. We have project templates that works fine without the ASP.NET Core reference [here](https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.ServiceBus.Topic/Dockerfile). I should ask @tomkerkhove for more info on this 🤔 .

[...]

PR #190, opened 2020-09-22
Remove the dependency to AspNetCore

fgheysels:
This PR is related to #189 

I've removed the dependencies to AspNetCore framework.

### DIFF
--- a/src/Arcus.Security.Core/Arcus.Security.Core.csproj
+++ b/src/Arcus.Security.Core/Arcus.Security.Core.csproj
@@ -17,17 +17,14 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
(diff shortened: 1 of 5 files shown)


fgheysels:
Woops, seems like I should've ran the testsuites before creating the PR :)

tomkerkhove:
That's why we have them here, no worries!

Thanks and LGTM other than the test.

fgheysels:
> That's why we have them here, no worries!
> 
> Thanks and LGTM other than the test.

I've updated the version of some packages, don't know if that might be an issue ? (2.2.0 > 3.1.8 for some extension packages)

tomkerkhove:
No problem at all, thanks @fgheysels!

tomkerkhove (approved):

fgheysels:
Thanks, now I can use Arcus.Security in a simple background service which runs as a container, on a base-image that does not contain the ASP.NET Core libraries / framework.

tomkerkhove:
Thank you!

PR #193, opened 2020-10-02
Feature - provide support for using new Azure.Identity SDK in Azure Key Vault secret provider

### LINKED ISSUE #191: Provide support for using Azure.Identity 📦 for Azure Key Vault
Provide support for using Azure.Identity 📦 for Azure Key Vault to support users who are already using that way of authenticating.

https://docs.microsoft.com/en-us/dotnet/azure/authentication

Thoughts?

stijnmoreels:
Provide capability to use both the 'old' as the 'new' SDK packages for interacting with the Azure Key Vault.
This includes double support in the secret provider itself, and updating the secret store builder extensions to have support for these new authentication types.

Closes #191

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
@@ -33,6 +33,7 @@ namespace Microsoft.Extensions.Hosting
         /// <param name="mutateSecretName">The optional function to mutate the secret name before looking it up.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> or <paramref name="certificate"/> is <c>null</c>.</exception>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="rawVaultUri"/> or <paramref name="clientId"/> is blank.</exception>
+        [Obsolete("Use the " + nameof(AddAzureKeyVaultWithCertificate) + " overload with the tenant ID instead")]
         public static SecretStoreBuilder AddAzureKeyVaultWithCertificate(
             this SecretStoreBuilder builder,
             string rawVaultUri,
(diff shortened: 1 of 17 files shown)


src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
```diff
@@ -263,7 +266,69 @@ public static class SecretStoreBuilderExtensions
...
+        /// <param name="configuration">The configuration related to the Azure Key Vault instance to use.</param>
+        /// <param name="cacheConfiguration">The configuration to control how the caching will be done.</param>
+        /// <param name="mutateSecretName">The optional function to mutate the secret name before looking it up.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/>, <paramref name="tokenCredential"/>, or <paramref name="configuration"/> is <c>null</c>.</exception>
+        public static SecretStoreBuilder AddAzureKeyVault(
+            this SecretStoreBuilder builder,
+            TokenCredential tokenCredential,
+            IKeyVaultConfiguration configuration,
+            ICacheConfiguration cacheConfiguration,
+            Func<string, string> mutateSecretName = null)
```
stijnmoreels, line 307: Added here a new extension with the base class `TokenCredential`; which is the base class for all the authentication types. (So service principal, managed identity, ...). Should we provide something simulair as to what we support or with the specific types (as in `ManagedIdentityCredential`), or something simulair like what we already have with the required values (ex. `clientId` and `clientSecret`) in the extension itself? We may have some conflicts there, though.

stijnmoreels: Didn't add docs just yet because it all depends on what we support here.

tomkerkhove: I personally would use both:
- We have the convenient extensions such as `AddAzureKeyVaultWithCertificate` which just take the required input. However, I would change the current ones to use Azure.Identity, not add new ones.
- We provide extensions to pass the TokenCredential directly in case they want to manage it.

tomkerkhove (commented):
Are you planning on writing integration tests for this as well?

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
stijnmoreels: > Are you planning on writing integration tests for this as well?

Well, if we would change the existing ones, there's not so much additional integration tests to be written. Except the one where we pass along the `TokenCredential` ourselves.

stijnmoreels: If we change the existing ones, we would have breaking changes because the `TokenCredential` implementations (such as `ClientCertificateCredential`) also takes in the `tenantId` which is now not currently available. Should I go through with the breaking change or should I mark them obsolete and create new ones.

tomkerkhove: > Should I go through with the breaking change or should I mark them obsolete and create new ones.

Let's add new ones and deprecate the current ones

stijnmoreels:
Hmm, not sure what I do wrong here. I created a new Managed Identity that is assigned to the used Key Vault but apparently all the tests fail bc the identity is _not_ assigned 🤔 .

stijnmoreels:
> Hmm, not sure what I do wrong here. I created a new Managed Identity that is assigned to the used Key Vault but apparently all the tests fail bc the identity is _not_ assigned 🤔 .

Ok, apparently it was already okay but I had to wait a little or so 😅.

src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -77,6 +84,44 @@ public KeyVaultSecretProvider(IKeyVaultAuthentication authentication, IKeyVaultC
                 "Requires the Azure Key Vault host to be in the right format, see https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#objects-identifiers-and-versioning");
 
             _authentication = authentication;
+            _isUsingAzureSdk = false;
```
tomkerkhove, line 87: Let's create an issue to remove our deprecated authentication in v2

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
```diff
@@ -86,6 +132,26 @@ public async Task KeyVaultSecretProvider_WithCustomManagedServiceIdentity_GetSec
             Assert.NotNull(secret.Version);
         }
 
+        [Fact]
+        public async Task KeyVaultSecretProvider_WithCustomManagedIdentity_GetSecret_Succeeds()
```
tomkerkhove, line 136: What od you mean with Custom Managed Identity? Do you mean User-assigned?

tomkerkhove (approved):
LGTM but added a few things

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
stijnmoreels: Yeah, meaning that we created a managed identity ourselves and didn't use the system identities.

src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: #202

src/Arcus.Security.Tests.Integration/KeyVault/KeyVaultSecretProviderTests.cs
tomkerkhove: Ok let's rename it to UserAssignedManagedIdentity then, that's how Azure calls it.

stijnmoreels: Done, resolving.

stijnmoreels:
Aha, @tomkerkhove , I think I know why the `ManagedIdentityCredential` doesn't work here; it's propably bc the application that calls it should be deployed on an Azure service or some sorts? Apparently, previous versions fell back on something else but this one doens't anymore.
As far from that, I don't know exactly why or what this exception should represent.

stijnmoreels:
> Aha, @tomkerkhove , I think I know why the `ManagedIdentityCredential` doesn't work here; it's propably bc the application that calls it should be deployed on an Azure service or some sorts? Apparently, previous versions fell back on something else but this one doens't anymore.
> As far from that, I don't know exactly why or what this exception should represent.

Should I then deploy an App Service that under the hood connects to Azure Key Vault and returns something so we're sure that the service can connect with managed identity to the Key Vault? 🤔

stijnmoreels:
@tomkerkhove , whoehoe, 🎉 🎉 all green.

tomkerkhove:
image

PR #198, opened 2020-10-11
Migrate from Polly-Signed  to Polly

### LINKED ISSUE #195: Migrate from `Polly-Signed` 📦 to `Polly` 📦
I'm using the Arcus.KeyVaultSecretProvider in a program that is running as a container in an Azure Container Instance.   This program uses Polly for retrying operations that have failed.

When my program starts, the container almost immediately stops.   The logs of the container contain this error message:

image

As it turns out, my program is using Polly 7.2.1 whereas the Arcus KeyVaultSecretProvider is using a package called `Polly-Signed` 5.9.0 which is build for .NET Standard 1.1 / .NET Framework 4.5 and dates back to April 2018.

I've now resolved the issue by also using this Polly-Signed package, so my container is now starting but I'm wondering if KeyVaultSecretProvider should also target the regular 'Polly' package ?  I'm also wondering how we can avoid such dependency conflicts in Arcus.

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

Closes #195

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
@@ -26,7 +26,7 @@
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
     <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="4.4.2" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.8" />
-    <PackageReference Include="Polly-Signed" Version="5.9.0" />
+    <PackageReference Include="Polly" Version="7.2.1" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
```diff
@@ -26,7 +26,7 @@
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
     <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="4.4.2" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.8" />
-    <PackageReference Include="Polly-Signed" Version="5.9.0" />
+    <PackageReference Include="Polly" Version="7.2.1" />
```
fgheysels, line 29: Just wondering if we should specify the version explicitly, or do something like `7.*` ?

fgheysels (approved):

stijnmoreels (approved):

src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
tomkerkhove: NuGet will allow for >7.2.1 but I presume you want this because you are on a lower version?

fgheysels: No, I'm not at a lower version.  
So, then it is fine for me I guess.

PR #201, opened 2020-10-19
Increase version of Microsoft.Azure.Services.AppAuthentication

### LINKED ISSUE #196: Using User assigned identity for KeyVault authentication requires newer version (>=1.4) of Microsoft.Azure.Services.AppAuthentication package
**Describe the bug**
When using a User Assigned Identity in an App Service for authenticating with KeyVault and retrieving a secret we get a KeyNotFoundException. Installing version 1.6.0 of Microsoft.Azure.Services.AppAuthentication resolved the issue.

**To Reproduce**

- Create an App Service
- Assign a user assigned identity which has secret get and list permissions on the KeyVault
- Setup Arcus with the KeyVault provider with connectionstring `RunAs=App;AppId={clientId of user assigned identity}`
- Use the ISecretprovider to retrieve a secret from the KeyVault
- The exception is thrown with message `(The secret <SecretName>was not found.) The secret <SecretName> was not found. None of the configured secret providers contains the requested secret`

**Expected behavior**
The app can access the secret

**Additional context**
Versions used
 - Arcus.Security.Providers.AzureKeyVault 1.3.0
 - Arcus.Security.Secrets.Core 0.4.0

After using a preview version with logging enabled I got the following error: `Unhandled exception. System.ArgumentException: Connection string RunAs=App;AppId={client id ellided} is not valid. Must contain 'TenantId' attribute and it must not be empty.`

I've found the following thread on StackOverflow describing the same issue: https://stackoverflow.com/questions/63608985/azure-app-service-with-user-assigned-managed-identity-crashes-application
The accepted answer suggests installing a newer version (>= 1.4) of the Microsoft.Azure.Services.AppAuthentication package. Installing 1.6.0 did indeed resolve the issue.

tomkerkhove:
Increase version of Microsoft.Azure.Services.AppAuthentication to fix authentication issues.

Closes #196

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
@@ -23,7 +23,7 @@
     <PackageReference Include="Microsoft.Azure.KeyVault" Version="3.0.2" />
     <PackageReference Include="Microsoft.Azure.KeyVault.Core" Version="3.0.1" />
     <PackageReference Include="Microsoft.Azure.KeyVault.WebKey" Version="3.0.2" />
-    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
+    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.5.0" />
     <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="4.4.2" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.8" />
     <PackageReference Include="Polly" Version="7.2.1" />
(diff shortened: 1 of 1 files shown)


src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
```diff
@@ -23,7 +23,7 @@
     <PackageReference Include="Microsoft.Azure.KeyVault" Version="3.0.2" />
     <PackageReference Include="Microsoft.Azure.KeyVault.Core" Version="3.0.1" />
     <PackageReference Include="Microsoft.Azure.KeyVault.WebKey" Version="3.0.2" />
-    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
+    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.5.0" />
```
stijnmoreels, line 26: Is this even something we keep supporting? I thought we would have to obsolete them, right?

tomkerkhove: Obsolete, yes but that doesn't mean we should stop supporting it as it doesn't work for now (see #196)

stijnmoreels:
OK, then, it's fine. Now only the build and we can merge!

stijnmoreels (approved):

tomkerkhove:
@stijnmoreels Had to update `Microsoft.IdentityModel.Clients.ActiveDirectory` as well

stijnmoreels (approved):

Issue #177, opened 2020-09-07
Log security event when requesting secret with `Arcus.Observability.Telemetry.Core` package once SQL package is removed


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
When the `CompositeSecretProvider` request a secret (via looking it up with the registered secret provider sources), we should log this as a security event with the `Arcus.Observability.Telemtry.Core` package since this is something we should filter upon when looking for security-related activities.

**Describe the solution you'd like**
Use the `ILogger.LogSecurityEvent` extension.

**Additional context**
Maybe we should wait until [this issue](https://github.com/arcus-azure/arcus.observability/issues/131) is done bc otherwise we'll get the SQL package dependency in our security package which is by all means not something we want.

tomkerkhove:
Can you elaborate a bit more what you want to achieve? You want to provide a security event for every single get secret operation per provider, is that correct?

stijnmoreels:
> Can you elaborate a bit more what you want to achieve? You want to provide a security event for every single get secret operation per provider, is that correct?

We have created the possibility to log security events with the `.LogSecurityEvent` in the `Arcus.Observability.Telemetry.Core` but that package has still a package reference to a SQL package. This issue is created to make sure when this SQL package is removed, we add the `Arcus.Observability.Telemetry.Core` package to the `Arcus.Security.Core` package and write a security event for every secret that is being requested.

PR #207, opened 2020-11-10
Feature - track Azure Key Vault dependency

### LINKED ISSUE #182: Provide capability to opt-in for dependency tracking for Azure Key Vault
Provide capability to opt-in for dependency tracking for Azure Key Vault

```csharp
// Adding the Azure Key Vault secret provider with the built-in overloads
builder.AddAzureKeyVaultWithManagedServiceIdentity(keyVaultUri, useDependencyTracking: true);
```

**Additional context**
We need to see if we can do this by bringing in that dependency or if that would cause circular references

stijnmoreels:
Opt-in to track the Azure Key Vault dependency, while maintaining backwards compatibility towards the preview v1.3.0 (which didn't have mutate secret names).

Closes #182 
Relates to #177

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Configuration/KeyVaultOptions.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Configuration/KeyVaultOptions.cs
@@ -1,0 +1,13 @@
+﻿namespace Arcus.Security.Providers.AzureKeyVault.Configuration
+{
+    /// <summary>
+    /// Represents the available options to configure extra options of the Azure Key Vault for the <see cref="KeyVaultSecretProvider"/>.
+    /// </summary>
+    public class KeyVaultOptions
+    {
+        /// <summary>
+        /// Gets or sets the flag to indicate whether or not the <see cref="KeyVaultSecretProvider"/> should track the Azure Key Vault dependency.
+        /// </summary>
+        public bool TrackDependency { get; set; } = false;
+    }
+}
--- a/src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
@@ -13,7 +13,6 @@ using Azure.Security.KeyVault.Secrets;
 using GuardNet;
 using Microsoft.Azure.KeyVault;
 using Microsoft.Azure.KeyVault.Models;
-using Microsoft.Rest.TransientFaultHandling;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Polly;
@@ -180,15 +179,13 @@ namespace Arcus.Security.Providers.AzureKeyVault
             Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Requires a non-blank secret name to request a secret in Azure Key Vault");
             Guard.For<FormatException>(() => !SecretNameRegex.IsMatch(secretName), "Requires a secret name in the correct format to request a secret in Azure Key Vault, see https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#objects-identifiers-and-versioning");
 
+            var isSuccessful = false;
             using (DependencyMeasurement measurement = DependencyMeasurement.Start())
             {
                 try
                 {
                     Secret secret = await GetSecretCoreAsync(secretName);
-                    if (_options.TrackDependency)
-                    {
-                        Logger.LogDependency("Azure key vault", secretName, VaultUri, isSuccessful: true, measurement: measurement); 
-                    }
+                    isSuccessful = true;
                     
                     return secret;
                 }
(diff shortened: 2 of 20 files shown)


src/Arcus.Security.Providers.AzureKeyVault/Configuration/KeyVaultOptions.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Arcus.Security.Providers.AzureKeyVault.Configuration
+{
+    /// <summary>
+    /// Represents the available options to configure extra options of the Azure Key Vault for the <see cref="KeyVaultSecretProvider"/>.
+    /// </summary>
+    public class KeyVaultOptions
```
tomkerkhove, line 6: Let's use the product name
```suggestion
    public class AzureKeyVaultOptions
```

src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -167,6 +175,31 @@ public virtual async Task<Secret> GetSecretAsync(string secretName)
...
+                    Logger.LogDependency("Azure key vault", secretName, VaultUri, isSuccessful: true, startTime: measurement.StartTime, duration: measurement.Elapsed);
+                    
+                    return secret;
+                }
+                catch
+                {
+                    Logger.LogDependency("Azure key vault", secretName, VaultUri, isSuccessful: false, startTime: measurement.StartTime, duration: measurement.Elapsed);
+                    throw;
+                }
+            }
```
tomkerkhove, line 198: This should simplify things a bit:
```suggestion
            bool isSuccessful = false;
            using (DependencyMeasurement measurement = DependencyMeasurement.Start())
            {
                try
                {
                    Secret secret = await GetSecretCoreAsync(secretName);
                    return secret;
                }
                finally
                {
	                if (_options.TrackDependency)
	                {
	                    Logger.LogDependency("Azure key vault", VaultUri, secretName, isSuccessful: isSuccessful, startTime: measurement.StartTime, duration: measurement.Elapsed);
                    }
                }
            }
```

I've swapped the vault & secret name as well, think this will be better for the rendered result.

Can you verify what it looks like please?

tomkerkhove (commented):
LGTM but made some suggestions.

Would you mind posting the result in application insights with these changes please?

src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: But we should also track the dependency when the secret retrieval was successful, of course. Otherwise, yeah. You're now also have additional `using` and `try/finally` logic, even when we don't need it, (that's what I was trying to avoid in the original one). But I guess this is also good.

stijnmoreels:
> LGTM but made some suggestions.
> 
> Would you mind posting the result in application insights with these changes please?

We could do that, I'll have to set up a Serilog logger so we can use our sink.

src/Arcus.Security.Providers.AzureKeyVault/Configuration/KeyVaultOptions.cs
stijnmoreels: You sure? We also have `KeyVaultConfiguration` and `KeyVaultSecretProvider`?

tomkerkhove: Then it's ok!

tomkerkhove: We should rename those for our next major version though

src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
tomkerkhove: It's not that the try/finally and using will make a difference :D

> But we should also track the dependency when the secret retrieval was successful, of course.

This is also included at the top sample, no?

tomkerkhove: My sample was not up-to-date, sorry; changed it to this:
image

stijnmoreels: Yep, is changed. Thanks.

stijnmoreels:
> LGTM but made some suggestions.
> 
> Would you mind posting the result in application insights with these changes please?

src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: Hmm, question, though, should we include the exception in the context of the failed depenency tracking?

tomkerkhove: Do you mean as dependency data or what do you mean? Do we have an overload for it?

stijnmoreels: Yeah, as dependency data or an entry in our telemetry context (`Dictionary<string, object>`). We don't have an overload for failures explicitly, no.

tomkerkhove (approved):

PR #208, opened 2020-11-12
Feature - opt-in of security events in secret store

### LINKED ISSUE #177: Log security event when requesting secret with `Arcus.Observability.Telemetry.Core` package once SQL package is removed

### LINKED ISSUE #205: Provide capability to opt-in for security events
Provide capability to opt-in for security events instead of turning them on by default as they generate a lot of telemetry which burries more relevant information.

stijnmoreels:
Provide the capability to opt-in for the emitted security events when requesting a secret through the secret store.
This includes the extension to configure this, alongside the necessary tests and docs update.

Closes #177 
Closes #205

### DIFF
--- a/src/Arcus.Security.Core/CompositeSecretProvider.cs
+++ b/src/Arcus.Security.Core/CompositeSecretProvider.cs
@@ -9,6 +9,7 @@ using GuardNet;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
 
 namespace Arcus.Security.Core
 {
@@ -19,6 +20,7 @@ namespace Arcus.Security.Core
     {
         private readonly IEnumerable<SecretStoreSource> _secretProviders;
         private readonly IEnumerable<CriticalExceptionFilter> _criticalExceptionFilters;
+        private readonly SecretStoreAuditingOptions _auditingOptions;
         private readonly ILogger _logger;
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


src/Arcus.Security.Core/CompositeSecretProvider.cs
```diff
@@ -225,14 +232,14 @@ private void EnsureAnySecretProvidersConfigured(string secretName)
...
+            if (_auditingOptions.EmitSecurityEvents)
             {
-                ["EventType"] = "Security",
-                ["SecretName"] = secretName,
-                ["SecretProviderType"] = source.SecretProvider.GetType().Name
-            });
+                _logger.LogSecurityEvent("Get Secret", new Dictionary<string, object>
+                {
+                    ["SecretName"] = secretName,
+                    ["SecretProviderType"] = source.SecretProvider.GetType().Name
```
tomkerkhove, line 240: Maybe just leave out the type as it will give the name or not? Would prefer a name over a type as a user
```suggestion
                    ["SecretProvider"] = source.SecretProvider.GetType().Name
```

tomkerkhove (approved):
Clean PR, thanks! LGTM

src/Arcus.Security.Core/CompositeSecretProvider.cs
stijnmoreels: We could maybe come up with an additional solution as wrapped instances (like cached) wil not show the inner type.

PR #210, opened 2020-11-17
Feature - provide capability to store in Azure Key Vault

tomkerkhove:
Provide the capability to store in Azure Key Vault.

Further improvements such as docs & integration with the cached provider will be done later on.

Relates to #206

PR #211, opened 2020-11-20
Feature - retrieve a specific secret provider based on a registered name

### LINKED ISSUE #204: Provide capability to assign name to providers added to the secret store
Provide capability to assign name to providers added to the secret store.

This way, we can inject the secret store in dependencies and request the secret provider with a given name to interact with only that secret store.

stijnmoreels:
Provides the capability to retrieve a specific `ISecretProvider`/`ICachedSecretProvider` based on a registered name. This happens when the provider is registered into the secret store (like the secret name mutation). An new interface is created `ISecretStore` to provide the functionality to retrieve these specific secret providers.

Closes #204


docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,68 @@
...
+Normally, you inject the `ISecretProvider` in your application service and accesses the secrets, but now you'll have to inject the `ISecretStore` interface to access the specific secret store operations.
+
+```csharp
+[ApiController]
+   public class OrderController : ControllerBase
+   {
+       public class OrderController(ISecretStore secretStore)
+       {
+            // Gets the `ISecretProvider` with the matched name.
+            ISecretProvider secretprovider = secretStore.GetProvider("get/set");
```
stijnmoreels, line 59: Should we allow this to return a typed instance instead so that the consumer doesn't have to cast it self?

stijnmoreels: It would be easier in combination with #210 .

tomkerkhove: I would provide both options so that people who care can specify it while others might not care

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,68 @@
...
+
+In those cases, you can register your secret provider with a unique name. This name will be matched with the secret provider so that in a later stage, you can retrieve back your named provider.
+
+## Register a named secret provider
+
+First, the secret provider has to be registered with a unique name. 
+Let's use this secret provider as an example:
+
+```csharp
+public class GetAndSetSecretprovider : ISecretProvider
```
tomkerkhove, line 19: I think in this case adding a custom provider does not bring a lot of added value and might confuse people, no? Maybe just show how this works with the existing providers?

src/Arcus.Security.Core/Extensions/SecretStoreBuilderExtensions.cs
```diff
@@ -30,7 +31,29 @@ public static class SecretStoreBuilderExtensions
...
+        /// <summary>
+        /// Adds a secret source to the secret store of the application that gets its secrets from the environment.
+        /// </summary>
+        /// <param name="builder">The builder to create the secret store.</param>
+        /// <param name="configureOptions">The function to configure the registration of the <see cref="ISecretProvider"/> in the secret store.</param>
+        /// <param name="target">The target on which the environment variables should be retrieved.</param>
+        /// <param name="prefix">The optional prefix which will be prepended to the secret name when retrieving environment variables.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="target"/> is outside the bounds of the enumeration.</exception>
+        public static SecretStoreBuilder AddEnvironmentVariables(
```
tomkerkhove, line 46: Personally I think it would be simpler if both name & mutation are just a parameter here instead of having ot use options, I don't know what @fgheysels thinks?

src/Arcus.Security.Providers.HashiCorp/Extensions/SecretStoreBuilderExtensions.cs
```diff
@@ -48,8 +81,49 @@ public static class SecretStoreBuilderExtensions
...
+        ///     or the <paramref name="secretPath"/> is blank.
+        /// </exception>
+        public static SecretStoreBuilder AddHashiCorpVaultWithUserPass(
+            this SecretStoreBuilder builder,
+            string vaultServerUriWithPort,
+            string username,
+            string password,
+            string secretPath,
+            Action<HashiCorpVaultUserPassOptions> configureOptions,
+            string name,
```
tomkerkhove, line 125: General remark, can't we make it the first parameter?

This is also how https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks does it

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
```diff
@@ -109,9 +111,10 @@ public static class SecretStoreBuilderExtensions
             string tenantId,
             string clientId,
             X509Certificate2 certificate,
-            bool allowCaching = false,
-            Func<string, string> mutateSecretName = null,
-            Action<KeyVaultOptions> configureOptions = null)
+            Action<KeyVaultOptions> configureOptions,
```
tomkerkhove, line 114: Can we move this to the end?

src/Arcus.Security.Providers.HashiCorp/Extensions/SecretStoreBuilderExtensions.cs
stijnmoreels: The name parameter is also the first one after the specific parameters of the secret provider itself. Same as here. Or am I missing something?

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
stijnmoreels: We could but we would mixing up specific with non-specific parameters. Not sure if that's a good idea?

docs/preview/features/secret-store/named-secret-providers.md
stijnmoreels: Done that. Resolving.

docs/preview/features/secret-store/named-secret-providers.md
stijnmoreels: Hmm, I thought with specifying a custom provider, it was more clear as of why you should use something that is already available in the 'default' secret store approach. 😅

src/Arcus.Security.Core/Extensions/SecretStoreBuilderExtensions.cs
stijnmoreels: I already updated this (as discussed by ourselves) to parameter-approach.

docs/preview/features/secret-store/named-secret-providers.md
tomkerkhove: Well I think they usecase is regardless of custom/built-in. You define all the secret stores that you use but in some cases you need a specific one because you have to interact with that store for example.

Example - I need a device authentication key which has a dedicated vault, so you request that specific secret provider

src/Arcus.Security.Providers.HashiCorp/Extensions/SecretStoreBuilderExtensions.cs
tomkerkhove: It's ok!

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
tomkerkhove: Can you elaborate on it please, not sure if I'm following!

docs/preview/features/secret-store/named-secret-providers.md
stijnmoreels: Aha, makes sense.

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
stijnmoreels: The 'specific' Azure Key Vault parameters (`TokenCredential`, `clientId`, `clientSecret`, `KeyVaultOptions`...) are parameters I would want first in the signature, before the 'general' parameters (`mutateSecretName`, `name`, (future) `optional`...).

Make sense?

docs/preview/features/secret-store/named-secret-providers.md
stijnmoreels: Used the built-in environment variable secret provider. Is that better?

tomkerkhove: Jup, thanks 👍

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
tomkerkhove: Fine by me!

stijnmoreels:
I think my part is done here? Can you approve if you agree @tomkerkhove ?

stijnmoreels:
@tomkerkhove , this is ready for your review.

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
+---
+title: "Named secret providers"
+layout: default
+---
+
+# Retrieve a specific secret provider from the secret store
+
+The default workings of the secret store, is that a set of secret providers are registered and the consumer gets access to all of secrets provided by `ISecretProvider`s.
```
tomkerkhove, line 8:
```suggestion
The default workings of the secret store, is that a set of secret providers are registered and the consumer gets access to all of secrets provided by using `ISecretProvider`.
```

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
...
+In some cases, you may want to retrieve a specific secret provider from the store. This can be because that secret provider has some functionality that the other providers doesn't have.
```
tomkerkhove, line 9:
```suggestion

In some cases, you may want to retrieve a specific secret provider from the store because that secret provider has some functionality that the other providers don't have.
```

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
...
+
+In those cases, you can register your secret provider with a unique name. This name will be matched with the secret provider so that in a later stage, you can retrieve back your named provider.
```
tomkerkhove, line 11:
```suggestion
In those cases, you can register your secret provider with a unique name so that, in a later stage, you can retrieve back your named provider.
```

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Register a named secret provider
```
tomkerkhove, line 13:
```suggestion
## Registering a named secret provider
```

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
...
+First, the secret provider has to be registered with a unique name. 
+
+```csharp
+.ConfigureSecretStore((config, stores) =>
+{
+    stores.AddEnvironmentVariables(..., name: "environment-variables");
+})
+```
+
+## Retrieve the named secret provider
```
tomkerkhove, line 25:
```suggestion
## Retrieving a named secret provider
```

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
...
+
+Now that the named environment variables secret provider is registered, we should be able to retrieve this provider somewhere else in the application.
```
tomkerkhove, line 27:
```suggestion
Now that the named environment variables secret provider is registered, we are able to retrieve this provider in our application.
```

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
...
+Normally, you inject the `ISecretProvider` in your application service and accesses the secrets, but now you'll have to inject the `ISecretStore` interface to access the specific secret store operations.
```
tomkerkhove, line 28:
```suggestion

Instead of injecting `ISecretProvider` in your application to access secrets, we'll inject `ISecretStore` interface to retrieve named secret providers.
```

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
...
+            var secretprovider = secretStore.GetProvider<EnvironmentVariableSecretProvider>("environment-variables");
+
+            // Gets the `ICachedSecretProvider` with the matched name (with either using the `ICachedSecretProvider` as return type or your own generic type).
+            // Mark that this only works when the secret provider was regisered as a cached secret provider.
+            ICachedSecretProvider cachedSecretProvider = secretStore.GetCachedProvider("your-cached-secret-provider");
+       }
+   }
+```
+
+> Note that the name of the registered `ISecretProvider` should be unique; otherwise and exception will be thrown when you try to access the `GetProvider` or `GetCachedProvider`.
```
tomkerkhove, line 46: Let's move this to where we register it
```suggestion
```

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -0,0 +1,46 @@
...
+Lets consider that you want explicitly use the built-in environment variables secret provider.
+First, the secret provider has to be registered with a unique name. 
+
+```csharp
+.ConfigureSecretStore((config, stores) =>
+{
+    stores.AddEnvironmentVariables(..., name: "environment-variables");
+})
+```
+
```
tomkerkhove, line 24:
```suggestion

⚠ The name of the registered secret providers should be unique; otherwise, an exception will be thrown when you try to access the `GetProvider` or `GetCachedProvider`.

```

tomkerkhove: However, for the last part I'd argue that the secret store builder should verify and block it so that it fails at startup, not runtime. Do you agree?

tomkerkhove (approved):
LGTM, just added some suggestions for the docs and improvement for name validation

docs/preview/features/secret-store/named-secret-providers.md
stijnmoreels: Yes, that would be better. I'll make sure we do that. Although, we can only verify it for the current secret store setup, not across setups. Let me clarify, we can't still guard against this:

```csharp
.ConfigureSecretStore(stores => stores.AddProvider(..., name: "duplicate-name"))
.ConfigureSecretStore(stores => stores.AddProvider(..., name: "duplicate-name"))
```

docs/preview/features/secret-store/named-secret-providers.md
stijnmoreels: Yes, used your suggestion. Resolving.

PR #216, opened 2020-12-02
Feature - add secret store to IServiceCollection

### LINKED ISSUE #214: How can I use the ISecretProvider if I do not want to use a HostBuilder
I've a project where I want to be able to configure and use one or multiple ISecretProviders, but I do not have (and I don't need) a HostBuilder instance.

In this scenario, how can I easily make use of the Arcus SecretProviders ?   I thought to do it like this:

```csharp
var secretProviderBuilder = new SecretStoreBuilder(services);

var secretSettings = config.GetSecretProviderSettings();

if (secretSettings.Type.Equals("KeyVault", StringComparison.OrdinalIgnoreCase))
{
    secretProviderBuilder.AddAzureKeyVaultWithManagedServiceIdentity(secretSettings.KeyVault.KeyVaultUrl);
}
else
{
   secretProviderBuilder.AddConfiguration(config.Configuration);
}

 var secretProvider =  services.BuildServiceProvider().GetService<ISecretProvider>();
```

But that doesn't work.  The SecretProviders are not registered in the ServiceCollection.   I've inspected the code and it turns out that I should call the `RegisterSecretStore` method on the `SecretStoreBuilder`, but that method is internal.

### LINKED ISSUE #215: Introduce `AddSecretStore` for configuring the store on `IServiceCollection`
Introduce `AddSecretStore` for configuring the store on services so that people don't have to rely on `HostBuilder`.

The same experience as we have with `HostBuilder`, but with `IServiceCollection`

stijnmoreels:
Adds the secret store to the `IServiceCollection` with an `.AddSecretStore` extension that adds the `ISecretProvider` component to the registered services, but then outside the .NET hosting.

Closes #214 
Closes #215

### DIFF
--- a/docs/preview/features/secret-store/index.md
+++ b/docs/preview/features/secret-store/index.md
@@ -83,23 +83,25 @@ The secret store is also available directly on the `IServiceCollection` for appl
 Just like you would register the secret store on the `HostBuilder`, you can use the `.AddSecretStore` extension to register the secret store:
 
 ```csharp
-IServiceCollection services = ...
-IConfiguration configuration = ...
-
-// Mark: the secret store uses logging, so make sure that you have set the logging beforehand.
-service.AddLogging(...);
-
-services.AddSecretStore(stores =>
+public void ConfigureServices(IServiceCollection services)
 {
-    stores.AddEnvironmentVariables();
-    
-    #if DEBUG
-    builder.AddConfiguration(configuration);
-    #endif
+    IConfiguration configuration = 
+        new ConfigurationBuilder()
+            .AddEnvironmentVariables()
+            .Build();
 
-    var keyVaultName = configuration["KeyVault_Name"];
-    stores.AddAzureKeyVaultWithManagedServiceIdentity($"https://{keyVaultName}.vault.azure.net");
-});
+    services.AddSecretStore(stores =>
+    {
+        stores.AddEnvironmentVariables();
+        
+        #if DEBUG
+        builder.AddConfiguration(configuration);
+        #endif
+    
+        var keyVaultName = configuration["KeyVault_Name"];
+        stores.AddAzureKeyVaultWithManagedServiceIdentity($"https://{keyVaultName}.vault.azure.net");
+    });
+}
 ```
 
 When the dependency injection container injects the dependent services in the rest of your application, 
--- a/docs/preview/features/secret-store/index.md
+++ b/docs/preview/features/secret-store/index.md
@@ -86,6 +86,9 @@ Just like you would register the secret store on the `HostBuilder`, you can use 
 IServiceCollection services = ...
 IConfiguration configuration = ...
 
+// Mark: the secret store uses logging, so make sure that you have set the logging beforehand.
+service.AddLogging(...);
+
 services.AddSecretStore(stores =>
 {
     stores.AddEnvironmentVariables();
--- a/docs/preview/features/secret-store/index.md
+++ b/docs/preview/features/secret-store/index.md
@@ -58,8 +58,7 @@ public class Program
                     builder.AddConfiguration(config);
 #endif
                     var keyVaultName = config["KeyVault_Name"];
-                    builder.AddEnvironmentVariables()
-                            .AddAzureKeyVaultWithManagedServiceIdentity($"https://{keyVaultName}.vault.azure.net");
+                    builder.AddAzureKeyVaultWithManagedServiceIdentity($"https://{keyVaultName}.vault.azure.net");
                 })
                 .ConfigureWebHostDefaults(webBuilder => webBuilder.UseStartup<Startup>());
     }
@@ -77,6 +76,31 @@ public class HealthController : ControllerBase
     }
 }
 ```
+
+### Using secret store outside .NET hosting
+The secret store is also available directly on the `IServiceCollection` for applications that run outside the .NET hosting but still want to use the Arcus secret store.
+
+Just like you would register the secret store on the `HostBuilder`, you can use the `.AddSecretStore` extension to register the secret store:
+
+```csharp
+IServiceCollection services = ...
+IConfiguration configuration = ...
+
+services.AddSecretStore(stores =>
+{
+    stores.AddEnvironmentVariables();
+    
+    #if DEBUG
+    builder.AddConfiguration(configuration);
+    #endif
+
+    var keyVaultName = configuration["KeyVault_Name"];
+    stores.AddAzureKeyVaultWithManagedServiceIdentity($"https://{keyVaultName}.vault.azure.net");
+});
+```
+
+When the dependency injection container injects the dependent services in the rest of your application, 
+the secret store will provide with you an `ISecretProvider` instance that contains the registered secret providers.
 
 ## Using secret store within Azure Functions
 
--- a/docs/preview/features/secret-store/index.md
+++ b/docs/preview/features/secret-store/index.md
@@ -53,12 +53,12 @@ public class Program
                 })
                 .ConfigureSecretStore((context, config, builder) =>
                 {
-                    builder.AddEnvironmentVariables();
 #if DEBUG
                     builder.AddConfiguration(config);
 #endif
                     var keyVaultName = config["KeyVault_Name"];
-                    builder.AddAzureKeyVaultWithManagedServiceIdentity($"https://{keyVaultName}.vault.azure.net");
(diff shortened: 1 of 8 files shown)


tomkerkhove:
@stijnmoreels is this ready for review or not? Because it's a draft PR but we're assigned so not sure what your expectations are!

stijnmoreels:
> @stijnmoreels is this ready for review or not? Because it's a draft PR but we're assigned so not sure what your expectations are!

Was still having a problem but fixed now. It's ready for review now! (Although I have a question).

docs/preview/features/secret-store/index.md
```diff
@@ -78,6 +77,34 @@ public class HealthController : ControllerBase
...
+The secret store is also available directly on the `IServiceCollection` for applications that run outside the .NET hosting but still want to use the Arcus secret store.
+
+Just like you would register the secret store on the `HostBuilder`, you can use the `.AddSecretStore` extension to register the secret store:
+
+```csharp
+IServiceCollection services = ...
+IConfiguration configuration = ...
+
+// Mark: the secret store uses logging, so make sure that you have set the logging beforehand.
+service.AddLogging(...);
```
stijnmoreels, line 90: Not sure what to do here. Should we make sure that the secret store can be set up without logging, or should we fail on missing logging infrastructure in the services?

src/Arcus.Security.Core/CompositeSecretProvider.cs
```diff
@@ -35,19 +35,18 @@ internal class CompositeSecretProvider : ICachedSecretProvider
         public CompositeSecretProvider(
             IEnumerable<SecretStoreSource> secretProviderSources, 
             IEnumerable<CriticalExceptionFilter> criticalExceptionFilters,
-            IOptions<SecretStoreAuditingOptions> auditingOptions,
```
stijnmoreels, line 38: Removed this (`IOptions<>`) as it's unnecessary complexity.

docs/preview/features/secret-store/index.md
fgheysels: My feelings are that you should not impede requirements / dependencies to the consumer of the library.   In other words, although logging is mostly required in projects, usage of the library should not fail if the consumer has not setup logging.

docs/preview/features/secret-store/index.md
```diff
@@ -78,6 +77,34 @@ public class HealthController : ControllerBase
...
+
+services.AddSecretStore(stores =>
```
fgheysels, line 92: `AddSecretStore` or `AddSecretProviders` ?

docs/preview/features/secret-store/index.md
fgheysels: I've talked about this with @tomkerkhove  a few days ago regarding logging in libraries.  Logging can be very handy here when troubleshooting, but in normal circumstances, I believe that logs from a library should be as concise as possible.   Libraries should log by default to a Trace / Verbose level imho.

docs/preview/features/secret-store/index.md
```diff
@@ -78,6 +77,34 @@ public class HealthController : ControllerBase
 }
 ```
 
+### Using secret store outside .NET hosting
+The secret store is also available directly on the `IServiceCollection` for applications that run outside the .NET hosting but still want to use the Arcus secret store.
```
fgheysels, line 81:
```suggestion
The secret store is also available directly on the `IServiceCollection` for applications that run without a .NET hosting context but still want to make use of the Arcus secret store.
```

docs/preview/features/secret-store/index.md
```diff
@@ -78,6 +77,34 @@ public class HealthController : ControllerBase
...
+
+Just like you would register the secret store on the `HostBuilder`, you can use the `.AddSecretStore` extension to register the secret store:
```
fgheysels, line 83:
```suggestion
Just like you would register the secret store on the `HostBuilder`, you can use the `.AddSecretStore` extension method to register the secret store:
```

docs/preview/features/secret-store/index.md
```diff
@@ -78,6 +77,34 @@ public class HealthController : ControllerBase
...
+
+```csharp
+IServiceCollection services = ...
+IConfiguration configuration = ...
```
fgheysels, line 87: Just asking, but why not 'completing' the code sample instead of omitting some things ?

fgheysels (commented):
Thanks for picking this up so quickly; would be really helpfull to have the ability to use Secret Providers without having a hosting context available

docs/preview/features/secret-store/index.md
tomkerkhove: Fully agree on both topics

docs/preview/features/secret-store/index.md
tomkerkhove: Let's stick with `AddSecretStore` to align with our `ConfigureSecretStore` unless you want to use `ConfigureSecretStore` here as well; but typically services use `Add` as far as I know

docs/preview/features/secret-store/index.md
```diff
@@ -58,8 +58,7 @@ public class Program
                     builder.AddConfiguration(config);
 #endif
                     var keyVaultName = config["KeyVault_Name"];
-                    builder.AddEnvironmentVariables()
```
tomkerkhove, line 61: Let's keep this version maybe?

docs/preview/features/secret-store/index.md
```diff
@@ -78,6 +77,34 @@ public class HealthController : ControllerBase
 }
 ```
 
+### Using secret store outside .NET hosting
```
tomkerkhove, line 80:
```suggestion
### Configuring secret store without .NET host builder
```

docs/preview/features/secret-store/index.md
tomkerkhove: @stijnmoreels we should change our secret store so that if there is no logger we use a nulllogger

docs/preview/features/secret-store/index.md
```diff
@@ -78,6 +77,34 @@ public class HealthController : ControllerBase
...
+    #if DEBUG
+    builder.AddConfiguration(configuration);
+    #endif
+
+    var keyVaultName = configuration["KeyVault_Name"];
+    stores.AddAzureKeyVaultWithManagedServiceIdentity($"https://{keyVaultName}.vault.azure.net");
+});
+```
+
+When the dependency injection container injects the dependent services in the rest of your application,
```
tomkerkhove, line 105: This is a bit too much implementation detail imo, what about this:
```suggestion
When your application wants to access a secret, all it has to do is use `ISecretProvider` which will give you access to all the registered secret providers.
```

What do you think?

docs/preview/features/secret-store/index.md
```diff
@@ -78,6 +77,34 @@ public class HealthController : ControllerBase
...
+    builder.AddConfiguration(configuration);
+    #endif
+
+    var keyVaultName = configuration["KeyVault_Name"];
+    stores.AddAzureKeyVaultWithManagedServiceIdentity($"https://{keyVaultName}.vault.azure.net");
+});
+```
+
+When the dependency injection container injects the dependent services in the rest of your application, 
+the secret store will provide with you an `ISecretProvider` instance that contains the registered secret providers.
```
tomkerkhove, line 106:
```suggestion
```

tomkerkhove (approved):

docs/preview/features/secret-store/index.md
tomkerkhove: This is maybe also something that should be opted-in/opted-out for in general, but we can discuss that in case that's a good idea

stijnmoreels: Yes, I also agree on this. I've changed it so logging is only used when it's available.

docs/preview/features/secret-store/index.md
stijnmoreels: Yeah, I added some extra code to make it more clear. Normally, I leave the 'general' types open to not confuse the sample.

docs/preview/features/secret-store/index.md
stijnmoreels: Yes, no problem!

docs/preview/features/secret-store/index.md
stijnmoreels: Yes, agree, keeping within the .NET naming.

Issue #217, opened 2020-12-03
How to use the Azure KeyVault SecretProvider with Managed service Identity


fgheysels:
**Description**
I am playing and digging around a bit with the Azure KeyVault SecretProvider (code that is now on t he master branch), and I'm wondering how I will be able to connect the SecretProvider to Azure KeyVault using Managed Service Identity.

I'm currently using a 1.4.0-alpha build and with that build, I'm able to connect to KeyVault using MSI like this:

```csharp
.ConfigureSecretStore((context, config, builder) =>
{
    builder.AddAzureKeyVaultWithManagedServiceIdentity(
                    securitySettings.KeyVault.KeyVaultUri, 
                    allowCaching: true,
                    mutateSecretName: secretName => secretName.Replace(":", "--"));
}
```

In the current version that is on master, it seems that this method has been marked as obsolete and you need to use the `AddAzureKeyVault
![image](https://user-images.githubusercontent.com/3605786/101094022-080b1580-35bc-11eb-8aee-66bd7c2ae901.png)

I don't have `clientId` however, I want to make use of Managed Service Identity.   
I see though that the `clientId` parameter of the proposed method is optional, 

![image](https://user-images.githubusercontent.com/3605786/101094280-6fc16080-35bc-11eb-9fbb-1e913f4b1385.png)

So I thought to give that a go.

![image](https://user-images.githubusercontent.com/3605786/101094390-a0a19580-35bc-11eb-9fe2-3a8421e027e9.png)

Unfortunately, this results in this exception:

```
Azure.Identity.AuthenticationFailedException : The ChainedTokenCredential failed to retrieve a token from the included credentials.
---- System.AggregateException : Multiple exceptions were encountered while attempting to authenticate. (ManagedIdentityCredential authentication unavailable. No Managed Identity endpoint found.) (EnvironmentCredential authentication unavailable. Environment variables are not fully configured.)
-------- Azure.Identity.CredentialUnavailableException : ManagedIdentityCredential authentication unavailable. No Managed Identity endpoint found.
-------- Azure.Identity.CredentialUnavailableException : EnvironmentCredential authentication unavailable. Environment variables are not fully configured.
   at Arcus.Security.Core.CompositeSecretProvider.WithSecretStoreAsync[T](String secretName, Func`2 callRegisteredProvider) in C:\Dev\R_and_D\arcus\arcus.security\src\Arcus.Security.Core\CompositeSecretProvider.cs:line 215
[...]

PR #218, opened 2020-12-07
Fix - use default Azure TokenCredential for more dev-friendliness

stijnmoreels:
Uses the default `DefaultAzureCredential` which behind the scenes uses the managed identity and environment variables (like we chained together), but also uses the Azure CLI information to make it more dev-friendly.

Relates to #217

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
@@ -326,7 +326,7 @@ namespace Microsoft.Extensions.Hosting
 
             return AddAzureKeyVault(
                 builder,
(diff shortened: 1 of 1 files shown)


src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
```diff
@@ -326,7 +326,7 @@ public static class SecretStoreBuilderExtensions
 
             return AddAzureKeyVault(
                 builder,
-                new ChainedTokenCredential(new ManagedIdentityCredential(clientId), new EnvironmentCredential()),
+                new DefaultAzureCredential(new DefaultAzureCredentialOptions { ManagedIdentityClientId = clientId }),
```
stijnmoreels, line 329: Is optional in options, so is correctly like this.

tomkerkhove (approved):

PR #220, opened 2020-12-15
Feature - simplify Azure Key Vault secret store overloads

### LINKED ISSUE #219: Be less demanding for Azure Key Vault registration
Be less demanding for Azure Key Vault registration because now it enforces to provide a name, secret name mutation & key vault options while people don't always need to use them.

They should be optional but because of colliding overloads I had to provide them.

**Additional context**
**Version:** 1.4.0-beta-3

stijnmoreels:
Provide a simplified version of each of the new SDK method overloads with only the caching as an additional option.
This lets consumers choose for a less complicated option when using the new SDK approach.

Closes #219

### DIFF
--- a/src/Arcus.Security.Tests.Integration/KeyVault/SecretStoreBuilderExtensionsTests.cs
+++ b/src/Arcus.Security.Tests.Integration/KeyVault/SecretStoreBuilderExtensionsTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


stijnmoreels:
The feature docs don't need to be updated bc it talks about the different arguments separately and we've only updated the extensions to a more simplified set.

src/Arcus.Security.Tests.Integration/KeyVault/SecretStoreBuilderExtensionsTests.cs
```diff
@@ -613,6 +614,42 @@ public async Task AddAzureKeyVaultWithTenant_WithServicePrincipal_GetSecretSucce
             Assert.NotNull(secret.Version);
         }
 
+        [Fact]
+        public async Task AddAzureKeyVaultWithTenant_WithServicePrincipal_GetSecretSucceeds()
```
stijnmoreels, line 618: I could be using an testing approach were we use functions as first class citizens, but that would only make it easier for seasoned devs, not for everyone, so for now this is probably the best way.

tomkerkhove:
Thanks!

With these changes, are we sure the current overloads are not longer colliding as well?

stijnmoreels:
> Thanks!
> 
> With these changes, are we sure the current overloads are not longer colliding as well?

Well, if you mean these new SDK overloads with the old ones, then there's no problem bc that was already looked into previously. The new SDK ones uses the client ID, token credential, .. which is unique for those ones.
If you mean with the other newly added overloads, then also no problem bc we didn't have used optional parameters.

tomkerkhove (approved):

PR #221, opened 2020-12-24
Feature - use '.AddSecretStore' in Azure Functions secret store extension

stijnmoreels:
Because of #215  , we are now able to use the `.AddSecretStore` directly on an `IServiceCollection` instance.
This can be used within the Azure Function extension too so we don't have to expose internal functionality.


tomkerkhove (approved):
Nice cleanup 👌

PR #222, opened 2020-12-24
Feature - use secret source name within security logging

stijnmoreels:
Since #204 we allow secret provider registrations to include a name that represents the secret source. This name can be used when we log a security even for the secret store.

### DIFF
--- a/src/Arcus.Security.Core/CompositeSecretProvider.cs
+++ b/src/Arcus.Security.Core/CompositeSecretProvider.cs
@@ -361,11 +361,13 @@ namespace Arcus.Security.Core
         {
             if (_auditingOptions.EmitSecurityEvents)
             {
-                _logger.LogSecurityEvent("Get Secret", new Dictionary<string, object>
+                var telemetryContext = new Dictionary<string, object>
                 {
                     ["SecretName"] = secretName,
-                    ["SecretProvider"] = source.SecretProvider.GetType().Name
-                }); 
+                    ["SecretProvider"] = source.Options.Name ?? source.SecretProvider.GetType().Name
+                };
+
+                _logger.LogSecurityEvent("Get Secret", telemetryContext);
             }
 
             Task<T> resultAsync = callRegisteredProvider(source);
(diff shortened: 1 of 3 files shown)


src/Arcus.Security.Core/CompositeSecretProvider.cs
```diff
@@ -364,8 +364,8 @@ private void EnsureAnySecretProvidersConfigured(string secretName)
                 _logger.LogSecurityEvent("Get Secret", new Dictionary<string, object>
                 {
                     ["SecretName"] = secretName,
-                    ["SecretProvider"] = source.SecretProvider.GetType().Name
-                }); 
+                    ["SecretProvider"] = source.Options.Name ?? source.SecretProvider.GetType().Name
```
stijnmoreels, line 367: We could of course add both the type and the name. So that the name would become `["SecretProviderName"]` or something.

src/Arcus.Security.Core/CompositeSecretProvider.cs
```diff
@@ -364,8 +364,8 @@ private void EnsureAnySecretProvidersConfigured(string secretName)
                 _logger.LogSecurityEvent("Get Secret", new Dictionary<string, object>
                 {
                     ["SecretName"] = secretName,
-                    ["SecretProvider"] = source.SecretProvider.GetType().Name
-                }); 
+                    ["SecretProvider"] = source.Options?.Name ?? source.SecretProvider.GetType().Name
```
tomkerkhove, line 367: Shouldn't we check for empty string?
```suggestion
                    ["SecretProvider"] = string.IsNullOrWhitespace(source.Options?.Name) ? source.SecretProvider.GetType().Name : source.Options?.Name
```

tomkerkhove (approved):

src/Arcus.Security.Core/CompositeSecretProvider.cs
stijnmoreels: The way we created these options, we already made sure that the model validates this. So, not sure, adding it would temper some paranoia fears but would also indicate that there's may not an validation happening on the options.

tomkerkhove: Ok, fine by me

PR #226, opened 2021-01-11
Feature - update to Arcus.Observability v2.0

stijnmoreels:
Updates to new version v2.0 of Arcus.Observability packae + making sure that we use the same package dependency strategy as the Observability library.

Closes #190

### DIFF
--- a/src/Arcus.Security.Core/Arcus.Security.Core.csproj
+++ b/src/Arcus.Security.Core/Arcus.Security.Core.csproj
@@ -17,15 +17,19 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="2.2.0" />
+  </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="3.1.8" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="3.1.8" />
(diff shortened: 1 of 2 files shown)


src/Arcus.Security.Core/Arcus.Security.Core.csproj
```diff
@@ -17,15 +17,19 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
stijnmoreels, line 20: We can do this bc it wasn't yet released.

tomkerkhove: Do we have to do this? Since it's the same package shouldn't we rely on NuGet to handle this?

stijnmoreels: This isn't about the Observability change, but about those other packages that were fixed on a v3.1.8 version; we changed that in observability and thought it will be a good thing to use that same approach here; splitting between netstandard and netcoreapp.

tomkerkhove: I'll leave it up to @fgheysels to decide

fgheysels: Dependencies, the pita in library development :)

- targetting a higher version of a dependent library means that consumers of Arcus must also use at least the version that is used by Arcus
- Does 3.1.8 not work on .NET Standard applications that do not target .NET Core 3.1 ?

stijnmoreels: I think it's mostly bc we support(ed) .NET Core 2.2 throughout Arcus (and still do).

tomkerkhove: If it's for .NET Core 2.2; ignore that since that's end-of-life

stijnmoreels: Ok, I change it back then to 3.1.8 packages.

PR #236, opened 2021-01-14
Feature - add Azure Key Vault secret store overloads w/o Managed Identity clientId

### LINKED ISSUE #230: Using AzureKeyVault secretprovider using Managed Identity always requires specifying a clientId in 1.4.0
I'm currently using Arcus.Security 1.4.0-beta-2 in certain projects and saw that 1.4.0 was released: Yaj!

When migrating from the beta-2 version to the offical release, I stumble upon some issues however.   I'm using this piece of code to use Azure KeyVault as a secretprovider:

```csharp
builder.ConfigureSecretStore(configure =>
{
    configure.AddAzureKeyVaultWithManagedIdentity(
        configuration["VesselSecretsKeyVault:Url"],
        allowCaching: true,
        mutateSecretName: n => n.Replace(":", "--"));
});
```

However, when migrating to 1.4.0, the above code no longer compiles.   It seems that in 1.4.0-beta-2, contains some optional parameters, for instance `clientId`.

image

In 1.4.0 however, `clientId` is mandatory to specify.   In most cases however, you want to make use of Managed Identity, so I actually don't want to specify the clientId:  this should be taken care of by Azure.

Please provide overloads of `AddAzureKeyVaultWithManagedIdentity` where a consumer doesn't need to specify the clientId so that the identity of the caller can be used.

stijnmoreels:
Adds some extra overloads for the Azure Key Vault secret provider in the secret store to authenticate with Managed Identity without a `clientId`.

Closes #230

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
@@ -377,6 +377,27 @@ namespace Microsoft.Extensions.Hosting
         /// </summary>
         /// <param name="builder">The builder to create the secret store.</param>
         /// <param name="rawVaultUri">The Uri of the Azure Key Vault you want to connect to.</param>
+        ///     The optional client id to authenticate for a user assigned managed identity.
+        ///     More information on user assigned managed identities can be found here: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#how-a-user-assigned-managed-identity-works-with-an-azure-vm</param>
+        /// <param name="allowCaching">The flag to indicate whether to include caching during secret retrieval in Azure key vault.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="rawVaultUri"/> is blank.</exception>
+        public static SecretStoreBuilder AddAzureKeyVaultWithManagedIdentity(
+            this SecretStoreBuilder builder,
+            string rawVaultUri,
+            bool allowCaching = false)
+        {
+            Guard.NotNull(builder, nameof(builder), "Requires a secret store builder to add the Azure Key Vault secret provider");
+            Guard.NotNullOrWhitespace(rawVaultUri, nameof(rawVaultUri), "Requires a non-blank URI of the Azure Key Vault instance to add the secret provider to the secret store");
+
+            return AddAzureKeyVaultWithManagedIdentity(builder, rawVaultUri, clientId: null, allowCaching: allowCaching);
+        }
+        
+        /// <summary>
+        /// Adds Azure Key Vault as a secret source which uses Managed Identity authentication.
+        /// </summary>
+        /// <param name="builder">The builder to create the secret store.</param>
+        /// <param name="rawVaultUri">The Uri of the Azure Key Vault you want to connect to.</param>
         /// <param name="clientId">
         ///     The optional client id to authenticate for a user assigned managed identity.
         ///     More information on user assigned managed identities can be found here: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#how-a-user-assigned-managed-identity-works-with-an-azure-vm</param>
@@ -399,6 +420,38 @@ namespace Microsoft.Extensions.Hosting
                 configureOptions: null,
                 name: null,
                 mutateSecretName: null,
+                allowCaching: allowCaching);
+        }
+
+        /// <summary>
+        /// Adds Azure Key Vault as a secret source which uses Managed Identity authentication.
+        /// </summary>
+        /// <param name="builder">The builder to create the secret store.</param>
+        /// <param name="rawVaultUri">The Uri of the Azure Key Vault you want to connect to.</param>
+        /// <param name="allowCaching">The flag to indicate whether to include caching during secret retrieval in Azure key vault.</param>
+        /// <param name="configureOptions">The optional additional options to configure the Azure Key Vault secret source.</param>
+        /// <param name="name">The unique name to register this Azure Key Vault provider in the secret store.</param>
+        /// <param name="mutateSecretName">The optional function to mutate the secret name before looking it up.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="rawVaultUri"/> is blank.</exception>
+        public static SecretStoreBuilder AddAzureKeyVaultWithManagedIdentity(
+            this SecretStoreBuilder builder,
+            string rawVaultUri,
+            Action<KeyVaultOptions> configureOptions,
+            string name,
+            Func<string, string> mutateSecretName,
+            bool allowCaching = false)
+        {
+            Guard.NotNull(builder, nameof(builder), "Requires a secret store builder to add the Azure Key Vault secret provider");
+            Guard.NotNullOrWhitespace(rawVaultUri, nameof(rawVaultUri), "Requires a non-blank URI of the Azure Key Vault instance to add the secret provider to the secret store");
+
+            return AddAzureKeyVaultWithManagedIdentity(
+                builder,
+                rawVaultUri,
+                clientId: null,
+                configureOptions: configureOptions,
+                name: name,
+                mutateSecretName: mutateSecretName,
                 allowCaching: allowCaching);
         }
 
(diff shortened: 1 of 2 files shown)


tomkerkhove (approved):

tomkerkhove:
@fgheysels Feel free to give this preview package onMyGet a spin if you can

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
```diff
@@ -372,6 +372,27 @@ public static class SecretStoreBuilderExtensions
...
+        /// <param name="rawVaultUri">The Uri of the Azure Key Vault you want to connect to.</param>
+        ///     The optional client id to authenticate for a user assigned managed identity.
+        ///     More information on user assigned managed identities can be found here: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#how-a-user-assigned-managed-identity-works-with-an-azure-vm</param>
+        /// <param name="allowCaching">The flag to indicate whether to include caching during secret retrieval in Azure key vault.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="rawVaultUri"/> is blank.</exception>
+        public static SecretStoreBuilder AddAzureKeyVaultWithManagedIdentity(
+            this SecretStoreBuilder builder,
+            string rawVaultUri,
+            bool allowCaching = false)
```
fgheysels, line 388: I know it is hard to find or decide which method overloads you should make.
I 'm just wondering if it wouldn't be helpfull to have the `mutateSecretName` parameter in this one as an optional argument ?

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
```diff
@@ -500,6 +553,29 @@ public static class SecretStoreBuilderExtensions
...
+        /// </summary>
+        /// <param name="builder">The builder to create the secret store.</param>
+        /// <param name="rawVaultUri">The Uri of the Azure Key Vault you want to connect to.</param>
+        /// <param name="cacheConfiguration">The configuration to control how the caching will be done.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="rawVaultUri"/> is blank.</exception>
+        public static SecretStoreBuilder AddAzureKeyVaultWithManagedIdentity(
+            this SecretStoreBuilder builder,
+            string rawVaultUri,
+            ICacheConfiguration cacheConfiguration)
```
fgheysels, line 567: I think it is sometimes confusing to have multiple overloads which actually offer the same functionality, but do it in a different way.

I am referring to this method which takes a `cacheConfiguration` argument, and the method that has an `allowCaching` parameter.
Maybe we should get rid of the method that takes an `allowCaching` parameter; if you want to have caching and are happy with the defaults, you could introduce `CacheConfiguration.Default` which you can pass to this method:

```csharp
builder.AddAzureKeyVaultWithManagedIdentity("myvaulturi", CacheConfiguration.Default);
```

fgheysels: Then, I think you could limit yourself to just one method for adding KeyVault using MSI, more specifically the one method that takes all the possible arguments, and specify defaults:

```csharp
AddAzureKeyVaultWithManagedIdentity(
    this SecretStoreBuilder builder,
    string vaultUri, // this one is mandatory offcourse
    string name = null,
    ICacheConfiguration cacheConfiguration = null, // no caching
    Func<string, string> mutateSecretName = null,
    Action<KeyVaultOptions> configureOptions);
```
Or maybe 2 methods.  You could just specify the overload which takes the one mandatory parameter as well, just for the sake of clarity ?

fgheysels (commented):
Thanks for the quick response on the issue I've opened :)

I've just some remarks which might improve the usability.

fgheysels:
> @fgheysels Feel free to give this preview package onMyGet a spin if you can

I'll do this next week.

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
stijnmoreels: > I think it is sometimes confusing to have multiple overloads which actually offer the same functionality, but do it in a different way.
> I am referring to this method which takes a cacheConfiguration argument, and the method that has an allowCaching parameter.
Maybe we should get rid of the method that takes an allowCaching parameter; if you want to have caching and are happy with the defaults, you could introduce CacheConfiguration.Default which you can pass to this method:

Yes, I see your point, but this is a suggestion outside the purpose of this PR. It's more a general concern since _all_ the overloads work like this (combination of `allowCaching` and `ICacheConfiguration`).

About the suggestion to make it optional parameters, I strongly suggest we don't do that bc that was the sole reason we had problems in the past with introducing new methods, signature conflicts, `MethodNotFound` failures... all bc it's harder to maintain to have optional parameters.
So, we've chosen to have method overloads with more or less parameters to choose from instead of one big method with almost everything optional.

fgheysels: I can agree with having multiple overloads instead of optionals in a library.  However, it might be fruitfull to re-think the overloads that are now being introduced to make sure that the most common scenario's are being covered without having to specify unnecessary parameters.
For instance, I think (now I'm speaking for myself, from my POV) the mutate-secretname parameter is one that is required in most cirumstances.  Also caching is often required (default would do).  I also wonder what is going on internally when you do not specify a name, does the library assign a name to the provider itself ?  (In that case, I do think name might be created as an optional argument, just like the allowCaching parameter is).

stijnmoreels: > it might be fruitfull to re-think the overloads that are now being introduced to make sure that the most common scenario's are being covered without having to specify unnecessary parameters.

Yes of course! Lets discuss this in a discussion/issue to see where we can go to. Having a default caching configuration instead of a boolean is something for example I'm all for.

> I also wonder what is going on internally when you do not specify a name, does the library assign a name to the provider itself ? 

If there's no name specified, then there's no name available internally too. It would be a kind off wrong design (IMO) to set a default name internally. Only 'named' secret secret providers which the consumer has configured, will have a name and is available when retrieving via the secret store.
And, yes, I can see the usefulness of optional parameters, but we try to limit here from using it bc it's a very hard thing to maintain. Also, we have created overloads with minimal and maximium parameters to circumvent this problem. 
But this can be discussed too in a different issue/disscussion.

stijnmoreels:
Is there any other comments related to this PR concrete or can we complete this?

fgheysels:
> @fgheysels Feel free to give this preview package onMyGet a spin if you can

Should I use 20210119.0.0-PR-236 or 20210114.0.0-PR-236 ?  Guess I'll take the latest one.

fgheysels:
I see that there are still some (non-obsolete) overloads of `AddAzureKeyVaultWithManagedIdentity` that take a `clientId` as parameter.  I wonder if this makes sense ?  In which situations are you going to have to specify the clientId yourself (without having to specify a corresponding client-secret ? ).

tomkerkhove:
That's required in case of user-defined managed identities AFAIK

fgheysels:
> That's required in case of user-defined managed identities AFAIK

Ah ok.  No experience with that so far.
It's not that this is somehow figured out by the underlying libraries ?  (I'd assume that for the authentication to KeyVault, it should not matter if the identity of the consumer is system assigned or user defined, but again, no experience with that).

fgheysels:
> Is there any other comments related to this PR concrete or can we complete this?

I think it is OK, except I find it a bit of a pity that there are 2 overloads where one method allows you to quickly use KeyVault with managed identity without having to need too much parameters.
The other method is the other end of the spectrum: all parameters need to be specified.   So, there's nothing in between.  

I can imagine  that there are scenario's where you want to specify the mutateSecretName parameter and the caching for instance, but none of all the others.
It would be a bit more developer friendly to have some additional overloads imho , but no hard requirement.

fgheysels (approved):

stijnmoreels:
> > Is there any other comments related to this PR concrete or can we complete this?
> 
> I think it is OK, except I find it a bit of a pity that there are 2 overloads where one method allows you to quickly use KeyVault with managed identity without having to need too much parameters.
> The other method is the other end of the spectrum: all parameters need to be specified. So, there's nothing in between.
> 
> I can imagine that there are scenario's where you want to specify the mutateSecretName parameter and the caching for instance, but none of all the others.
> It would be a bit more developer friendly to have some additional overloads imho , but no hard requirement.

Yes, I understand, also something we can discuss somewhere else.

PR #245, opened 2021-01-19
Feature - support subset of secret providers in secret store

### LINKED ISSUE #237: Registering multiple SecretProviders that are grouped together
It is currently possible to register multiple SecretProviders.    When retrieving a secret, the chain of registered SecretProviders is called, and the secret is retrieved (cfr how the Configuration in .NET works).
It is currently also possible to register a SecretProvider with a specific name so that this specific SecretProvider can be retrieved from the SecretStore.

Wouldn't it be nice if you could combine those 2 features together ?   That is: registering multiple SecretProviders with the same name, so that they would form a 'group' of secretproviders ?   You can then retrieve that group of secretproviders from the SecretStore via the name, just like you can retrieve a single secretprovider by name now.

I don't think the API needs to change for implementing this feature.  I think, behind the scenes a dictionary can be used where the 'key' is the SecretProvider-Name and the value contains a CompositeSecretProvider where all SecretProviders that are registered with the same name are added to.

stijnmoreels:
Add support for registering multiple `ISecretProvider` implementations with the same name so they are combined in a subset upon retrieval.

Closes #237

### DIFF
--- a/docs/preview/features/secret-store/named-secret-providers.md
+++ b/docs/preview/features/secret-store/named-secret-providers.md
@@ -7,9 +7,9 @@ layout: default
 
 The default workings of the secret store, is that a set of secret providers are registered and the consumer gets access to all of secrets provided by using `ISecretProvider`.
 
-In some cases, you may want to retrieve a specific secret provider from the store because that secret provider has some functionality that the other providers don't have.
+In some cases, you may want to retrieve a specific secret provider or a subset of secret providers from the store because that secret provider has some functionality that the other providers don't have.
 
-In those cases, you can register your secret provider with a unique name so that, in a later stage, you can retrieve back your named provider.
+In those cases, you can register your secret provider(s) with a name so that, in a later stage, you can retrieve back your named provider(s).
 
 ## Registering a named secret provider
 
@@ -22,8 +22,6 @@ First, the secret provider has to be registered with a unique name.
     stores.AddEnvironmentVariables(..., name: "environment-variables");
 })
 ```
-
-⚠ The name of the registered secret providers should be unique; otherwise, an exception will be thrown when you try to access the `GetProvider` or `GetCachedProvider`.
 
 ## Retrieving a named secret provider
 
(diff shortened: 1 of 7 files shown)


stijnmoreels:
Should hold of until #236 is done and v1.4.1 is released bc this one is for v1.5.

src/Arcus.Security.Core/CompositeSecretProvider.cs
```diff
@@ -127,70 +139,74 @@ public ISecretProvider GetProvider(string name)
...
-                throw new KeyNotFoundException(
-                    $"Could not retrieve the named {nameof(ISecretProvider)} because more than one provider was registered with the name '{name}'");
+                return subsetSecretStore.Value;
             }
-
-            SecretStoreSource firstProvider = matchingProviders.First();
-            return firstProvider;
+            
+            throw new KeyNotFoundException(
+                $"Could not retrieve the named {nameof(ISecretProvider)} because no secret provider was registered with the name '{name}'");
```
stijnmoreels, line 209:
```suggestion

            throw new KeyNotFoundException(
                $"Could not retrieve the named {nameof(ISecretProvider)} because no secret provider was registered with the name '{name}'");
```

src/Arcus.Security.Core/ISecretStore.cs
```diff
@@ -51,12 +46,13 @@ public interface ISecretStore
...
+        /// <exception cref="KeyNotFoundException">Thrown when there was no <see cref="ICachedSecretProvider"/> found in the secret store with the given <paramref name="name"/>.</exception>
+        /// <exception cref="NotSupportedException">
+        ///     Thrown when their was either none of the registered secret providers are registered as <see cref="ICachedSecretProvider"/> instances
+        ///     or there was an <see cref="ISecretProvider"/> registered but not with caching.
         /// </exception>
-        /// <exception cref="NotSupportedException">Thrown when there was an <see cref="ICachedSecretProvider"/> registered but not with caching.</exception>
         /// <exception cref="InvalidCastException">Thrown when the registered <see cref="ICachedSecretProvider"/> cannot be cast to the specific <typeparamref name="TCachedSecretProvider"/>.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when multiple <see cref="ICachedSecretProvider"/> were registered with the same name.</exception>
         TCachedSecretProvider GetCachedProvider<TCachedSecretProvider>(string name) where TCachedSecretProvider : ICachedSecretProvider;
```
fgheysels, line 56: Why do we need the explicit `GetCachedProvider` method ?  Is the consumer who wants to retrieve a secretprovider interested in the fact if the secret provider is cached or not ?   In what situation would one prefer to call this method instead of the more general `GetProvider` method ?

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -38,13 +36,14 @@ using Arcus.Security.Core.Providers;
...
 {
-
     [ApiController]
     public class OrderController : ControllerBase
     {
         public class OrderController(ISecretStore secretStore)
         {
              // Gets the `ISecretProvider` with the matched name (with either using the `ISecretProvider` as return type or your own generic type).
+             // ⚠ The name of the registered secret providers should be unique when retrieving the concrete secret provider; 
+             //         otherwise, an exception will be thrown when you try to access the `GetProvider<>` or `GetCachedProvider<>`.
```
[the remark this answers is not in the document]
stijnmoreels: Yes, I understand your confusion, but it's still correct.
When you use either `.GetProvider<TConcreteSecretProvider>` or `.GetCachedSecretProvider<TConcreteSecretProvider>`, you still need to have a unique name, because the targetted secret provider will be cast to the concrete implementation.

When multiple providers have the same name, in this case, we can't cast it.

src/Arcus.Security.Core/ISecretStore.cs
```diff
@@ -51,12 +46,13 @@ public interface ISecretStore
...
-        ///     or there were multiple <see cref="ICachedSecretProvider"/> instances registered with the same name.
+        /// <exception cref="KeyNotFoundException">Thrown when there was no <see cref="ICachedSecretProvider"/> found in the secret store with the given <paramref name="name"/>.</exception>
+        /// <exception cref="NotSupportedException">
+        ///     Thrown when their was either none of the registered secret providers are registered as <see cref="ICachedSecretProvider"/> instances
+        ///     or there was an <see cref="ISecretProvider"/> registered but not with caching.
         /// </exception>
-        /// <exception cref="NotSupportedException">Thrown when there was an <see cref="ICachedSecretProvider"/> registered but not with caching.</exception>
         /// <exception cref="InvalidCastException">Thrown when the registered <see cref="ICachedSecretProvider"/> cannot be cast to the specific <typeparamref name="TCachedSecretProvider"/>.</exception>
+        /// <exception cref="InvalidOperationException">Thrown when multiple <see cref="ICachedSecretProvider"/> were registered with the same name.</exception>
```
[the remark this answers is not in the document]
stijnmoreels: 1. It's thrown (as explained above), because when accessing a specific concrete secret provider, we will have to cast it, and multiple registrations will result in a composite, so we can't cast it.
2. Yes, of course, you can register multiple cached secret providers with the same name too, and retrieve them with `.GetCachedSecretProvider`. But if you want an specific instance, you'll have to specify it with a unique name.

src/Arcus.Security.Core/ISecretStore.cs
stijnmoreels: The reason here why we have both the 'general' `.GetProvider` and the 'concrete' `.GetProvider<TScretProvider>`; is because people may want to use specific operations on the concrete implementation. The Azure Key Vault secret provider has for example the possibility to store secrets too. But that functionality can't be accessed through the 'general' `ISecretProvider` interface. So we will have to cast that.

In case of a cached version, the same. When people implement their own version, they may want to access specific functionality from that implementation that's not available in the `ISecretProvider`.

docs/preview/features/secret-store/named-secret-providers.md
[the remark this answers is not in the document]
fgheysels: Are there use cases in which were we want or need to specify the concrete type when retrieving the provider ?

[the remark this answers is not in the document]
stijnmoreels: Yes, when storing Azure Key Vault secrets. That's available in the secret provider, but not in the 'general' `ISecretProvider` interface.

docs/preview/features/secret-store/named-secret-providers.md
```diff
@@ -7,9 +7,9 @@ layout: default
 
 The default workings of the secret store, is that a set of secret providers are registered and the consumer gets access to all of secrets provided by using `ISecretProvider`.
 
-In some cases, you may want to retrieve a specific secret provider from the store because that secret provider has some functionality that the other providers don't have.
+In some cases, you may want to retrieve a specific secret provider or a subset of secret providers from the store because that secret provider has some functionality that the other providers don't have.
 
-In those cases, you can register your secret provider with a unique name so that, in a later stage, you can retrieve back your named provider.
+In those cases, you can register your secret provider(s) with a name so that, in a later stage, you can retrieve back your named provider(s).
```
fgheysels, line 12:
```suggestion
In those cases, you can register your secret provider(s) with a name so that, in a later stage, you can retrieve your named provider(s).
```

src/Arcus.Security.Core/ISecretStore.cs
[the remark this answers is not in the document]
fgheysels: Why would I use `GetCachedSecretProvider` instead of `GetSecretProvider<ICachedSecretProvider>` ?  I mean, what's the added value of this method ?

src/Arcus.Security.Core/CompositeSecretProvider.cs
```diff
@@ -100,25 +107,30 @@ internal class CompositeSecretProvider : ICachedSecretProvider, ISecretStore
             {
                 return concreteProvider;
             }
+            
+            if (provider is CompositeSecretProvider)
+            {
+                throw new InvalidOperationException(
+                    $"Cannot cast registered {nameof(ISecretProvider)} with name '{name}' to type '{typeof(TSecretProvider).Name}' " +
+                    $"because multiple secret providers were registered with the name '{name}'");
```
fgheysels, line 115: Maybe an idea to include in the exception message how this can be solved ?  Maybe mention that you should use the non generic method to retrieve the secretproviders if multiple secretproviders have been registered with the same name ?

For instance:
>"Multiple SecretProviders have been registered with name {name}.  Use the non-generic GetProvider method to retrieve them"

stijnmoreels:
> Why would I use GetCachedSecretProvider instead of GetSecretProvider<ICachedSecretProvider> ? I mean, what's the added value of this method ?

For example, if you register the Azure Key Vault secret provider with caching and you want to store a secret with the provider, you'll have to use `.GetCachedProvider<KeyVaultCachedSecretProvider>`.

src/Arcus.Security.Core/CompositeSecretProvider.cs
stijnmoreels: Yes, thanks, did that. Resolving.

fgheysels (approved):

tomkerkhove (approved):

PR #247, opened 2021-01-22
🚧 Feature - use default `CacheConfiguration` instead of `allowCaching`

### LINKED ISSUE #239: Improve how caching should be specified when registering a secret provider
Right now, there are 2 ways to specify that caching must be used when registering a `ISecretProvider`.
One approach is to specify `allowCaching = true` when registering a SecretProvider.  Another approach is to specify an `ICacheConfiguration`.

These are 2 different approaches to serve the same functionality.   Wouldn't it be nicer to have only one of those options available ?

I would suggest to get rid of the methods which take the `allowCaching` parameter so that you can only specify an `ICacheConfiguration` when you want to enable caching.
If the default caching configuration suffices for a consumer, he should be able to specify caching like this:

```csharp
builder.AddAzureKeyVaultWithManagedIdentity("myvaulturi", CacheConfiguration.Default);
```

(See also my remark on this PR: #236 )

stijnmoreels:
Use the `CacheConfiguration.Default` instead of `allowCaching` in the secret store overloads to add Azure Key Vault as a secret provider to the secret store.

Closes #239

### DIFF
--- a/docs/preview/features/secret-store/provider/key-vault.md
+++ b/docs/preview/features/secret-store/provider/key-vault.md
@@ -17,6 +17,7 @@ PM > Install-Package Arcus.Security.Providers.AzureKeyVault
 After installing the package, the addtional extensions becomes available when building the secret store.
 
 ```csharp
+using Arcus.Security.Core.Caching.Configuration;
 using Microsoft.Extensions.Hosting;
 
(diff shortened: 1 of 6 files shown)


stijnmoreels:
Close for now.

PR #251, opened 2021-02-11
Feature - add security policy to the repo

stijnmoreels:
Makes sure that users have a option when creating issues to create a security vulnerability.

The details, as of what they should do, is probably something we should discuss. But having at least the file in place is the first step, I guess.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
+reported vulnerability, what to expect if the vulnerability is accepted or
+declined, etc.


SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
```
tomkerkhove, line 13: These don't really look correct?

tomkerkhove (changes requested):

SECURITY.md
stijnmoreels: It's more like a template, like we have the issue template. So there's some guidance to add a security vulnerability.

tomkerkhove: Not really, it's the security policy of our project so that users can see what our security support model is.

For example https://github.com/tomkerkhove/promitor/security and shows up in the issue template chooser:
image


If this is for reporting security vulnerabilities, we should use an issue template instead similar to https://github.com/tomkerkhove/promitor/issues/new?assignees=tomkerkhove&labels=security&template=3_security_issue.yml (this is using a rendered issue preview so you can ignore that, but you get the point)

stijnmoreels: Aha, OK, we should consider adding both then, but then I'm not sure how this version table works exactly.

tomkerkhove: If we want to yes, but it's a commitment.

The table says which versions we will patch in case of security vulnerabilities.

Issue #262, opened 2021-03-08
[Docs] Remove obsolete Azure Key Vault documentation from current documentation


stijnmoreels:
**Describe what the problem is**
Apparently, we have still a obsolete feature documentation on authentication with Azure Key Vault that uses old/obsolete features. This was previously required before the introduction of the secret store but should now be removed from the current version feature docs.
https://security.arcus-azure.net/features/auth/azure-key-vault

This was first discovered by @pim-simons.

PR #264, opened 2021-03-10
Fix - broken link to managed identity auth in Azure Key Vault

stijnmoreels:
Let's try to add a dedicated file that redirects to the new Azure Key Vault managed identity authentication.

Relates to #262


stijnmoreels:
> What is the goal of this PR?

1. Trying different way to redirect the deprecated page, since the one we tried failed.
2. Removing the links to the deprecated page from the index pages in the current version

tomkerkhove (approved):

stijnmoreels:
Hmm, I still think it doesn't work. Maybe I'll just add a page with a link that it's moved so that you manually choose the redirect.

PR #282, opened 2021-06-29
feat: open HashiCorp Vault secret provider for extension

stijnmoreels:
Make the `HashiCorpSecretProvider` open for extension by virtualizing the public secret provider's methods and give access to private members for child implementations.

Relates to https://github.com/arcus-azure/arcus/issues/143


docs/preview/features/secret-store/provider/hashicorp-vault.md
```diff
@@ -130,4 +130,114 @@ public class Program
 }
 ```
 
+### Custom implementation
```
tomkerkhove, line 133: All looks good to me, I just don't know if we have to add all of this documentation, what do you think @fgheysels ?

tomkerkhove (approved):

docs/preview/features/secret-store/provider/hashicorp-vault.md
stijnmoreels: We can maybe minimize it. Only show them in the overloads but not a fully working example like now. Would that be better?

docs/preview/features/secret-store/provider/hashicorp-vault.md
```diff
@@ -130,4 +130,114 @@ public class Program
...
+    {
+        await _retryPolicy.ExecuteAsync(async () => 
+        {
+            await base.GetRawSecretAsync(secretName);
+        });
+    }
+}
+```
+
+As you can see, we allo both secret retrieval methods to be overwritten so we can prepend our retryable functionality.
```
fgheysels, line 177:
```suggestion
As you can see, we allow both secret retrieval methods to be overwritten so we can prepend our retryable functionality.
```
Typo :)

fgheysels:
```suggestion
As you can see, we allo both secret retrieval methods to be overridden so we can prepend our retryable functionality.
```
overwritten or overridden ?

fgheysels:
```suggestion
As you can see, we allo both secret retrieval methods to be overwritten so we can prepend our retry functionality.
```

docs/preview/features/secret-store/provider/hashicorp-vault.md
fgheysels: I think it's good to have thorough documentation :)
Drawback is offcourse that you also need to maintain it, but I'm a fan of this kind of docs.   It's clear and explains what must be done.

It's also a good idea to mention the fact that you can create a custom extension method for using your custom type, but, putting the example code for that might be a bit too much.   You could also refer to a source file which defines this extension method for the providers that we provide.

src/Arcus.Security.Providers.HashiCorp/HashiCorpSecretProvider.cs
```diff
@@ -53,24 +51,42 @@ public class HashiCorpSecretProvider : ISecretProvider
...
         /// Retrieves the secret value, based on the given name
         /// </summary>
         /// <param name="secretName">The name of the secret key</param>
         /// <returns>Returns the secret key.</returns>
-        /// <exception cref="ArgumentException">The <paramref name="secretName"/> must not be empty</exception>
-        /// <exception cref="ArgumentNullException">The <paramref name="secretName"/> must not be null</exception>
-        /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
-        public async Task<string> GetRawSecretAsync(string secretName)
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="secretName"/> is blank.</exception>
+        public virtual async Task<string> GetRawSecretAsync(string secretName)
```
fgheysels, line 86: I'm still more a fan off dedicated / restricted extension points, but that's maybe just me.
For instance OnBefore / OnAfter overrideable methods.
Or, a default retry-mechanisms, that can be modified by the consumer.   

With making the public methods virtual, you bring off course more power to t he consumer (and more responsabilities).

stijnmoreels: > I'm still more a fan off dedicated / restricted extension points, but that's maybe just me.

I see that 😄 , I think if we look for a broader approach for all Arcus types, it could be quite a job to maintain such methods as people may want to overwrite different things. Especially if you handle return values, alternate inputs... it would also become (I think) quite difficult to know where exactly the overridden call gets executed (I always have that problem when I come across such implementations).
But yeah, that could just be me 😄 .

> With making the public methods virtual, you bring off course more power to t he consumer (and more responsabilities).

True, that's very true. With this, we allow more functionality to be changed, but it's also an approach we can use for every Arcus type, making it more user-friendly in my opinion. We have already received multiple requests for this, that they could override the methods. So, it seems like a logical choice.

But, if you have an alternate example of how this would work, and how we can streamline it across possible other types, feel free to show us!

tomkerkhove: > I'm still more a fan off dedicated / restricted extension points, but that's maybe just me.
> For instance OnBefore / OnAfter overrideable methods.
> Or, a default retry-mechanisms, that can be modified by the consumer.

I can see cases where that's useful indeed, but I mainly see those as "we allow you to do things, but the core is solely for us since it's kind of tricky and we shouldn't mess up".

While in the majority of the cases, you just want to extend what happens - But I'm fine with either, these would just add more methods but that's fine I guess.

stijnmoreels:
What can be the conclusion here? @tomkerkhove Can I go through and merge or should there be something different. We can open a discussion where we could explore alternatives, but for now I think it's okay.

src/Arcus.Security.Providers.HashiCorp/HashiCorpSecretProvider.cs
tomkerkhove: With regards to https://github.com/arcus-azure/arcus.security/pull/282#issuecomment-878787196, I'd ask @fgheysels instead of me @stijnmoreels

docs/preview/features/secret-store/provider/hashicorp-vault.md
tomkerkhove: Sounds good, let's keep it as is

src/Arcus.Security.Providers.HashiCorp/HashiCorpSecretProvider.cs
fgheysels: Actually, I do acknowledge that there are pro's and con's for both approaches:

- Controlled extension points:  might be enough for 80% of the users, but it is possible that you don't cover the other 20%of the cases.
- Public virtual:  you give tremendous freedom to the user of the library, but it might be easier for the consumer to mess up.  On the other hand: we do not need to take developers by the hand.   Disadvantage:  once you do this, there's no way back.

I don't feel that this is up to me / us to decide.  A discussion with users of the Arcus library or a survey can shed some more light on what the 'Arcus customers' actually want.

fgheysels:
> What can be the conclusion here? @fgheysels Can I go through and merge or should there be something different. We can open a discussion where we could explore alternatives, but for now I think it's okay.
> This secret provider is the same way overridable as the `KeyVaultSecretProvider`, so in any case, it should be streamlined.

I don't want to be the blocker if I'm the only one with a different opinion :)
 I still think it 's good to have a small inquiery with other users, but off course, that takes time ....    

@tomkerkhove  what is your opinion ?

ps: [this](https://softwareengineering.stackexchange.com/questions/396959/never-make-public-members-virtual-abstract-really) is also a valuable discussion on the topic.

tomkerkhove:
I'd suggest to move with this as-is and continue the overall discussion on https://github.com/arcus-azure/arcus/issues/143 and go through all libs based on that outcome

PR #310, opened 2022-02-08
fix: Improve the obsolete-warning regarding default caching

fgheysels:
The warning message that was displayed to developers when they used the obsolete version of `AddAzureKeyVaultWithManagedServiceIdentity` was displayed as:

> Use the AddAzureKeyVaultWithServicePrincipal method overload with Default instead to allow default caching.

Seems like the `nameof(CacheConfiguration.Default)` just gets translated to `Default` and this is a bit confusing.

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
@@ -384,7 +384,7 @@ namespace Microsoft.Extensions.Hosting
         /// <param name="allowCaching">The flag to indicate whether to include caching during secret retrieval in Azure key vault.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c>.</exception>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="rawVaultUri"/> is blank.</exception>
-        [Obsolete("Use the " + nameof(AddAzureKeyVaultWithManagedIdentity) + " method overload with " + nameof(CacheConfiguration.Default) + " instead to allow default caching")]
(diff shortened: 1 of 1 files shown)


src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
```diff
@@ -384,7 +384,7 @@ public static class SecretStoreBuilderExtensions
         /// <param name="allowCaching">The flag to indicate whether to include caching during secret retrieval in Azure key vault.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="builder"/> is <c>null</c>.</exception>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="rawVaultUri"/> is blank.</exception>
-        [Obsolete("Use the " + nameof(AddAzureKeyVaultWithManagedIdentity) + " method overload with " + nameof(CacheConfiguration.Default) + " instead to allow default caching")]
+        [Obsolete("Use the " + nameof(AddAzureKeyVaultWithManagedIdentity) + " method overload and pass CacheConfiguration.Default as parameter for the cacheConfiguration to allow default caching")]
```
stijnmoreels, line 387:
```suggestion
        [Obsolete("Use the " + nameof(AddAzureKeyVaultWithManagedIdentity) + " method overload and pass " + nameof(CacheConfiguration) + "." + nameof(CacheConfiguration.Default) + " as parameter for the cacheConfiguration to allow default caching")]
```

I would be more in favor of doing it this way, to make sure that the docs are always linked to the code.

stijnmoreels (commented):
Small be great improvement. Thx! 👍

src/Arcus.Security.Providers.AzureKeyVault/Extensions/SecretStoreBuilderExtensions.cs
fgheysels: Indeed, that is better.   I didn't think of it to do it like this.

PR #318, opened 2022-02-18
chore: correct net6.0 nuget packages dependencies

stijnmoreels:
Reference dependencies based on target framework. Uses new Observability package when using .NET 6.

Relates to https://github.com/arcus-azure/arcus/issues/188


fgheysels (approved):
LGTM.
I think it would be good if we could already use a pre-release version of this, together with the latest observability packages

PR #321, opened 2022-02-18
docs: add dedicate page for using secret store within Azure Functions

### LINKED ISSUE #301: Add extra link on secret store index page to 'How to use secret store in Azure Functions'
> We should have some simple documentation on how to use / integrate the ISecretProvider in an Azure Function project.

>I know that we consider the Arcus Templates as documentation / getting started as well, but I think it is not sufficient in this case.
(For instance, people that already have an existing Function who want to make use of the SecretProvider, or if you started a Function project but didn't use the template because you're not going to use a Http trigger (AFAIK, there only exists a template for functions which creates a Function that is triggered via HTTP).

We should add an extra link to the secret store index page for more visibility.

stijnmoreels:
Adds a dedicated page to inform users about the use of the Arcus secret store within Azure Functions. Also adds a link in the general index page so users can be directed directly to it.

Closes #301


docs/preview/features/secret-store/azure-functions.md
```diff
@@ -0,0 +1,64 @@
...
+
+namespace MyHttpAzureFunction
+{
+    public class Startup : FunctionsStartup
+    {
+        public override void Configure(IFunctionsHostBuilder builder)
+        {
+            builder.ConfigureSecretStore(stores =>
+            {
+                stores.AddEnvironmentVariables();
```
fgheysels, line 31: I don't think we need this here, as this is also added on line 34

docs/preview/features/secret-store/azure-functions.md
```diff
@@ -0,0 +1,64 @@
...
+
+                var keyVaultName = config["KeyVault_Name"];
```
fgheysels, line 33: This code sample gives a nice overview, but should we -for completeness- also show where `config` is coming from by adding a ctor and injecting `IConfiguation` into the `Startup` class ?

fgheysels (commented):
Thanks, just a few remarks

docs/preview/features/secret-store/azure-functions.md
stijnmoreels: Ah yes, that's correct. Let's remove this.

docs/preview/features/secret-store/azure-functions.md
stijnmoreels: We should probably use `IFunctionsHostBuilder.GetContext().Configuration` for this. Yes, let's add that.

fgheysels (approved):

PR #326, opened 2022-02-21
docs: update readme with available secret providers

stijnmoreels:
The README file wasn't up-to-date. This fixes that.


README.md
```diff
@@ -9,24 +9,37 @@ Security for Azure development in a breeze.
...
-- **HashiCorp Vault secret provider**
-
-```shell
-PM > Install-Package Arcus.Security.Providers.HashiCorp
-```
-
 - **Secret store**
 
 ```shell
 PM > Install-Package Arcus.Security.Core
```
fgheysels, line 15: Actually, if you install a package for a secret-provider, this one is implicitly installed as well if I'm not mistaken.
Are there cases where you want to install this package separately ?

fgheysels (approved):

README.md
stijnmoreels: Hmm, I think of places where you want to use `ISecretProvider` with your own implementation, or just want to be dependent on the secret store. I think we used the README to give an overview of all the packages, but not sure if we do that for every repo and if we should do it like this.

fgheysels: Then we might need to add a short explanation on what that package does or not ?

stijnmoreels: Yes, that would be better. 👍

stijnmoreels: Done that and also added links to the docs site. https://github.com/arcus-azure/arcus.security/pull/326/commits/b4d8349f19d42e19edcad0475e464d6dbd3ae0b1

fgheysels (approved):

Issue #317, opened 2022-02-18
Remove Azure Key Vault secret and vault name validation from Azure Key Vault secret provider


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
As a continuation of https://github.com/arcus-azure/arcus.observability/issues/287, we should also remove the Azure Key Vault validation on the public methods of the Azure Key Vault secret provider.

**Describe the solution you'd like**
Remove all the guard regex validations on the Azure Key Vault secret provider, so it solely verify for a non-blank secret name.

**Additional context**
Original discussion: https://github.com/arcus-azure/arcus.observability/discussions/260

PR #328, opened 2022-02-24
chore: correct marking az keyvault secret name obsolete

stijnmoreels:
In previous PR, https://github.com/arcus-azure/arcus.security/pull/324, we marked the dependency name as obsolete, while this was not correct. Only the regular expressions and the patterns should be marked as obsolete.

Relates to #317

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
+++ b/src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
@@ -39,6 +39,7 @@ namespace Arcus.Security.Providers.AzureKeyVault
         /// <summary>
         /// Gets the pattern which a Azure Key Vault secret name should match against. (See https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#objects-identifiers-and-versioning).
         /// </summary>
+        [Obsolete("Will be removed in v2.0")]
         internal const string SecretNamePattern = "^[a-zA-Z][a-zA-Z0-9\\-]{0,126}$";
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
```diff
@@ -28,7 +28,6 @@ public class KeyVaultSecretProvider : ISecretProvider
         /// <summary>
         /// Gets the name of the dependency that can be used to track the Azure Key Vault resource in Application Insights.
         /// </summary>
-        [Obsolete("Will be removed in v2.0")]
         protected const string DependencyName = "Azure key vault";
```
fgheysels, line 31:
```suggestion
        protected const string DependencyName = "Azure Key Vault";
```
?

fgheysels: woops :)

fgheysels (approved):

src/Arcus.Security.Providers.AzureKeyVault/KeyVaultSecretProvider.cs
stijnmoreels: Ah, no, this is not something we can change or control. I think this is a 'special' name that Azure Application Insights uses to correctly assign the right Azure resource.

stijnmoreels: Yes, this affect the Azure Application Insights dependency call. If we don't use this name, it will show up differently. Not recognized as a Azure Key Vault dependency. This was how it was when we developed dependency tracking.

stijnmoreels: This is for example a set of constants of an AppInsights package, to give you an idea: https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/WEB/Src/DependencyCollector/DependencyCollector/Implementation/RemoteDependencyConstants.cs

fgheysels: Thanks for these insights!

PR #333, opened 2022-03-08
feat: add config az func secret store extension

### LINKED ISSUE #334: Add missing extension for registering secret store in Azure Functions
Currently, we have an `.ConfigureSecretStore` extension for Azure Functions to register the secret store, but when using the `IConfiguration` and/or the context (environment information), one still has to call and retrieve it themselves via `IFunctionsHostBuilder.GetContext`.

As we provide an extension overload in general applications with the `IConfiguration`, we should consider to streamline this in Azure Functions as well. We can use the `FunctionsHostBuilderContext` and `IConfiguration` besides the `SecretStoreBuilder` in the overload.

stijnmoreels:
Adds an `.ConfigureSecretStore` to the Azure Functions environment that has access to the `IConfiguration` and `IFunctionsHostContext`. This extra method overload will streamline the secret store registration with the already available extensions on `IHostBuilder`.

Closes #334


src/Arcus.Security.AzureFunctions/Extensions/IFunctionHostBuilderExtensions.cs
```diff
@@ -27,5 +28,23 @@ public static IFunctionsHostBuilder ConfigureSecretStore(this IFunctionsHostBuil
...
+        /// <param name="configureSecretStores">The customization of the different target secret store sources to include in the final <see cref="ISecretProvider"/>.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="functionsHostBuilder"/> or <paramref name="configureSecretStores"/> is <c>null</c>.</exception>
+        public static IFunctionsHostBuilder ConfigureSecretStore(
+            this IFunctionsHostBuilder functionsHostBuilder,
+            Action<FunctionsHostBuilderContext, IConfiguration, SecretStoreBuilder> configureSecretStores)
+        {
+            Guard.NotNull(functionsHostBuilder, nameof(functionsHostBuilder), "Requires a functions host builder to add the secret store");
+            Guard.NotNull(configureSecretStores, nameof(configureSecretStores), "Requires a function to configure the secret store with potential secret providers");
+
+            FunctionsHostBuilderContext context = functionsHostBuilder.GetContext();
```
stijnmoreels, line 45: Not sure if this should be called here or inside the `AddSecretStore`. The call would be delayed, and outside of the context. I didn't see much context-specifics when inspecting `.GetContext` but not sure if we should take the risk of moving it inside the lambda.

fgheysels: I would keep it as is.

docs/preview/features/secret-store/azure-functions.md
```diff
@@ -28,8 +28,7 @@ namespace MyHttpAzureFunction
     {
         public override void Configure(IFunctionsHostBuilder builder)
         {
-            IConfiguration config = builder.GetContext().Configuration;
-            builder.ConfigureSecretStore(stores =>
+            builder.ConfigureSecretStore((FunctionsHostBuilderContext context, IConfiguration config, SecretStoreBuilder stores) =>
```
fgheysels, line 31: Technically , this is a breaking change...

We must also make sure that this change i reflected in our docs / code samples.

fgheysels (approved):

docs/preview/features/secret-store/azure-functions.md
stijnmoreels: Ah, I think not, bc we have only added an overload. So you could still use the one from before, it's just that there's now a new one that is 'better' so we could update the docs.

fgheysels: Shall we update the documentation in this PR as well ?

stijnmoreels: The documentation is already updated, this comment is even on a updated preview docs. 😉

fgheysels: I need to pay more attention :D

fgheysels (approved):

PR #338, opened 2022-03-18
docs: publish v1.7 feature docs

stijnmoreels:
Publishes the v1.7 feature docs.


docs/versioned_docs/version-v1.7.0/features/secrets/general.md
```diff
@@ -0,0 +1,72 @@
...
+var cacheConfiguration = new CacheConfiguration(TimeSpan.FromMinutes(10)); // Optional: Default is 5 min
+var cachedSecretProvider = new CachedSecretProvider(secretProvider, cacheConfiguration);
+Secret secret = await cachedSecretProvider.GetSecretAsync("EventGrid-AuthKey");
+```
+
+## Forcing a secret refresh
+In some scenarios you'd like to skip the cache and do a hard refresh by looking it up in the provider.
+
+This is important because in certain scenarios your secrets can be rolled and thus you will be revoked access.
+After a hard refresh you can use the latest secret again and proceed your work.
```
fgheysels, line 58: When reading this sentence, I'd assume that you need to call a specific function to do a 'hard refresh', but in fact, it is just calling the `GetSecretAsync` method with the `ignoreCache` parameter set to true.

Therefore, I'd change this sentence to something like:
> When you want to bypass the cached secret, call the `GetSecretAsync` method and specify `true` for the `ignoreCache` parameter.

Then I also wonder, does this parameter have a side-effect of invalidating the cache for this specific secret ?

stijnmoreels: We have a 'InvalidateSecretAsync' method on our cached secret providers that will do the 'hard refresh' that is described here.

stijnmoreels: > Then I also wonder, does this parameter have a side-effect of invalidating the cache for this specific secret ?

This is a parameter on an interface that is publicly available to be used alongside the existing set of secret providers, so consumers can have this done by their own choosing. We do not, in fact, cause side-effects. We only by-pass the caching. https://github.com/arcus-azure/arcus.security/blob/aba283e79ce68408752a1c7ad3b0ec0db43ec979/src/Arcus.Security.Core/Caching/CachedSecretProvider.cs#L144

fgheysels: I would just make this more clear, because the code sample does not align with 'forcing a a secret refresh'.
Changing the title to 'bypassing cached secrets' would improve the docs a lot imho.

fgheysels: I would just make this more clear, because the code sample does not align with 'forcing a a secret refresh'.
Changing the title to 'bypassing cached secrets' would improve the docs a lot imho.

docs/versioned_docs/version-v1.7.0/features/secrets/general.md
```diff
@@ -0,0 +1,72 @@
...
+By default, retrieved secrets are cached for **5 minutes**, but you can configure this yourself.
+
+```csharp
+var cacheConfiguration = new CacheConfiguration(TimeSpan.FromMinutes(10)); // Optional: Default is 5 min
+var cachedSecretProvider = new CachedSecretProvider(secretProvider, cacheConfiguration);
+Secret secret = await cachedSecretProvider.GetSecretAsync("EventGrid-AuthKey");
+```
+
+## Forcing a secret refresh
+In some scenarios you'd like to skip the cache and do a hard refresh by looking it up in the provider.
```
fgheysels, line 55:
```suggestion
In some scenarios you'd like to skip the cache and retrieve the secret by looking it up in the secret-store, instead of retrieving it from the cache.
```

docs/versioned_docs/version-v1.7.0/features/secrets/general.md
fgheysels, line 58: This line 58 fits more in the section regarding invalidating the cache.

docs/versioned_docs/version-v1.7.0/features/secrets/general.md
stijnmoreels: We should change this in all versions, then.

docs/versioned_docs/version-v1.7.0/features/secrets/general.md
stijnmoreels: Ah, yes, but we have to change this in all versions, then.

stijnmoreels:
I would suggest to do all these suggestions on all the other versions in a diff PR, bc the longer we wait to complete this one, the longer consumers don't have any docs on the latest version.

- [ ] Update to better sub-title 'forcing secret refresh' to 'by-pass'
- [ ] Remove 'force secret refresh' in sentence
- [ ] Move line 58 to 'by-pass

fgheysels:
> I would suggest to do all these suggestions on all the other versions in a diff PR, bc the longer we wait to complete this one, the longer consumers don't have any docs on the latest version.
> 
> * [ ]  Update to better sub-title 'forcing secret refresh' to 'by-pass'
> * [ ]  Remove 'force secret refresh' in sentence
> * [ ]  Move line 58 to 'by-pass

Fine for me, although I don't know if it takes that much extra time ?

stijnmoreels:
Well, if we need to update all versions. 😄 , since my priorities are somewhere else now.

docs/versioned_docs/version-v1.7.0/features/secrets/general.md
fgheysels: Do you see a problem in that ?

stijnmoreels: No, only will take a bit of time. Will do it probably later, after all the other more critical stuff is done. So, not in this PR.

stijnmoreels:
Ok, I'll complete this and we can later improve it #340 .

PR #339, opened 2022-03-18
chore: update with az keyvault as default package

stijnmoreels:
Since we stepped off from the `Arcus.Security.All` package, let's use a different 'default' package in the release notes.

### DIFF
--- a/build/nuget-release.yml
+++ b/build/nuget-release.yml
@@ -118,8 +118,8 @@ stages:
             parameters:
               repositoryName: 'arcus-azure/arcus.security'
               releaseNotes: |
-                Install new version via [NuGet](https://www.nuget.org/packages/Arcus.Security.All/$(Build.BuildNumber))
+                Install new version via [NuGet](https://www.nuget.org/packages/Arcus.Security.Providers.AzureKeyVault/$(Build.BuildNumber))
                 ```shell
-                PM > Install-Package Arcus.Security.All --Version $(Build.BuildNumber)
+                PM > Install-Package Arcus.Security.Providers.AzureKeyVault --Version $(Build.BuildNumber)
                 ```
           - template: nuget/publish-official-package.yml@templates


fgheysels (approved):

build/nuget-release.yml
```diff
@@ -118,8 +118,8 @@ stages:
             parameters:
               repositoryName: 'arcus-azure/arcus.security'
               releaseNotes: |
-                Install new version via [NuGet](https://www.nuget.org/packages/Arcus.Security.All/$(Build.BuildNumber))
+                Install new version via [NuGet](https://www.nuget.org/packages/Arcus.Security.Providers.AzureKeyVault/$(Build.BuildNumber))
```
fgheysels, line 121: I do wonder if we still need this task ?
As we are also calling the 'publish-official-packages' task, which pubishes nuget packages 

image

I think it doesn't make sense to say 'install a new version via nuget' with a link to a 'random' default package.  Arcus.Security consists of a set of packages, and people need to decide themselves which packages are required.

Or, we should list them all here.

```
Install a new version of Arcus.Security.Core via [NuGet](link to Arcus.Security.Core)
Install a new version of Arcus.Security.AzureKeyVault via NuGet
etc...
```

Or, just a generic description that says:
> Install a new version of the Arcus.Security package that you want via NuGet

fgheysels: (I think this is true for other Arcus packages as well).

fgheysels (approved):

build/nuget-release.yml
stijnmoreels: This is the GitHub release task, not the NuGet task. 
We could indeed be more detailed here instead of only describing one package. Though, the release notes will maybe a bit bloated if we do that.

stijnmoreels: It is true that this is something we do across all Arcus libraries. So, if we plan on changing, we should track this on the general repo to change all libraries.

fgheysels: I just think it makes no sense to mention a 'random' package from this project .

stijnmoreels: It made more sense, maybe, when we still had that 'all' package.
I'll create an issue for this on the general repo, so we can discus it there.

stijnmoreels: Added here: https://github.com/arcus-azure/arcus/issues/198

fgheysels: I'll resolve this so that we can complete this, as it also makes no sense to currently refer to a non existing or unlisted package

fgheysels (approved):
Let's complete this, and improve it via another PR.

PR #358, opened 2022-08-23
feat: add secret version support

### LINKED ISSUE #354: Take into account secondary (or more) secrets during secret retrieval
When a secret is rotated, usually the application uses secondary secrets to keep everything available. The secret store, though, is only configurable for single-secret retrieval while in those cases multiple secrets should be taken into account.

We should be able to get multiple secrets based on a single secret name. This can be done by configuring this single-to-multiple configuration during the registration of the secret provider. The consumer should then determine that (for example) secrets with `MySecret` should result in `MySecret-Primary` and `MySecret-Secondary`. 

Something in the like of:
```csharp
services.AddSecretStore(stores =>
{
    stores.AddAzureKeyVault(...);

    stores.AddDimensionalSecretName("MySecret", secretName =>
    {
        return new [] { secretName + "-Primary", secretName + "-Secondary"  });
    })
});
```

With an extension on the `ISecretProvider`, we could provide an easy access so that you could call `.GetRawSecretsAsync()` for multiple secrets, if available.

**Additional context**
Follow-up issue discussion/proposal of https://github.com/arcus-azure/arcus.webapi/issues/358
@gverstraete , any ideas, remarks?

stijnmoreels:
Add support for registering multiple secret versions for a secret name on secret provider registrations.

Closes #354

### DIFF
--- a/docs/preview/03-Features/secret-store/create-new-secret-provider.md
+++ b/docs/preview/03-Features/secret-store/create-new-secret-provider.md
@@ -1,4 +1,4 @@
----
+﻿---
 title: "Create custom secret provider"
 layout: default
 ---
@@ -203,6 +203,8 @@ public class RegistrySecretProvider : IVersionedSecretProvider
 
 The `amountOfVersions` can be configured via the secret provider options (`.AddVersionedSecret`).
 Each secret provider registration has the ability to register a amount of secret versions for secret name, that amount is passed to your implementation. For more information, see [this dedicated page](./versioned-secret-provider.md).
+
+> 💡 Note that versioned secrets can be combined with caching. The set of secrets will be cached, just like a single secret.
 
 ### Adding secret name mutation before looking up secret
 When you want secret names 'changed' or 'mutated' before they go through your secret provider (ex. changing `Arcus.Secret` to `ARCUS_SECRET`);
--- a/docs/preview/03-Features/secret-store/create-new-secret-provider.md
+++ b/docs/preview/03-Features/secret-store/create-new-secret-provider.md
@@ -167,7 +167,7 @@ namespace Application.Controllers
 
 ### Adding secret versions to your secret provider
 When your secret storage location supports secret versions, you could consider adapting your secret provider to support these secret versions.
-For more information on how you can use many secret versions in your application, see [this dedicated page](./versioned-secret-provider.md).
+For more information on how you can use multiple versions of a secret in your application, see [this dedicated page](./versioned-secret-provider.md).
 
 Implement from `IVersionedSecretProvider` instead of `ISecretProvider` to allow the secret store to pick that your secret provider supports secret versions.
 The following example shows how the registry secret provider only supports two versions of a secret:
--- a/docs/preview/03-Features/secret-store/create-new-secret-provider.md
+++ b/docs/preview/03-Features/secret-store/create-new-secret-provider.md
@@ -166,7 +166,7 @@ namespace Application.Controllers
 ```
 
 ### Adding secret versions to your secret provider
-When your secret storage location supports secret versions, you could consider adapting your secret provider to support these secret versions.
+When your secret storage location supports versioned secrets, you could consider adapting your secret provider to support these.
 For more information on how you can use multiple versions of a secret in your application, see [this dedicated page](./versioned-secret-provider.md).
 
 Implement from `IVersionedSecretProvider` instead of `ISecretProvider` to allow the secret store to pick that your secret provider supports secret versions.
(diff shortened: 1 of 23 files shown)


stijnmoreels:
Was already working on this on and off. Think this is everything needed, but required another set of eyes.

docs/preview/03-Features/secret-store/create-new-secret-provider.md
```diff
@@ -181,8 +165,48 @@ namespace Application.Controllers
 }
 ```
 
-### Adding secret name mutation before looking up secret
+### Adding secret versions to your secret provider
+When your secret storage location supports secret versions, you could consider adapting your secret provider to support these secret versions.
```
fgheysels, line 169:
```suggestion
When your secret storage location supports versioned secrets, you could consider adapting your secret provider to support retrieving secrets by version.
```
Maybe this makes it a bit more clear ?

docs/preview/03-Features/secret-store/create-new-secret-provider.md
```diff
@@ -181,8 +165,48 @@ namespace Application.Controllers
...
+For more information on how you can use many secret versions in your application, see [this dedicated page](./versioned-secret-provider.md).
```
fgheysels, line 170:
```suggestion
For more information on how you can use multiple versions of a secret in your application, see [this dedicated page](./versioned-secret-provider.md).
```

docs/preview/03-Features/secret-store/create-new-secret-provider.md
```diff
@@ -181,8 +165,48 @@ namespace Application.Controllers
...
+The following example shows how the registry secret provider only supports two versions of a secret:
+```csharp
+using Arcus.Security.Core;
+using Microsoft.Win32;
+
+public class RegistrySecretProvider : IVersionedSecretProvider
+{
+    // Also implement the general `ISecretProvider` methods...
+
+     public Task<IEnumerable<string>> GetRawSecretsAsync(string secretName, int amountOfVersions)
```
fgheysels, line 182: So, you retrieve actually a 'history' of secrets .   Wouldn't it be better that you can define:  Hey, I want version X of this secret ?

fgheysels: Hmm, I might get the idea.  You want to get for instance the last 3 values of a certain secret, and if the most recent version doesn't work, you fall back on the previous one ?

stijnmoreels: > So, you retrieve actually a 'history' of secrets . Wouldn't it be better that you can define: Hey, I want version X of this secret ?

That was not the initial reason why this feature was introduced. The purpose was to make sure that multiple versions of a single secret (name) were able to be retrieved from the secret store. This would allow (for example) easier key rotation in Key Vault while still supporting 'older' versions, up to the one the consumer defines.
From a user perspective, (if the secret versions are generated, for example), that would be an implementation detail and of no use to the actual secret value.

> Hmm, I might get the idea. You want to get for instance the last 3 values of a certain secret, and if the most recent version doesn't work, you fall back on the previous one ?

Yes, implementations that uses the secret store and use it like that. Depends on how the secret store is used and application specifics, of course.

docs/preview/03-Features/secret-store/create-new-secret-provider.md
stijnmoreels: It's not 'by version' that they are requested, but by amount.
`When your secret storage location supports versioned secrets, you could consider adapting your secret provider to support retrieving multiple versions of a secret.`

stijnmoreels: Changed in https://github.com/arcus-azure/arcus.security/pull/358/commits/b9fbf2f85da1d9a0d5c7cac73ceae752a90966dc

docs/preview/03-Features/secret-store/create-new-secret-provider.md
fgheysels: Yes, I agree.  Because typically, consumers don't care about the version-number.   Read the issue / feat request as well in the meantime :)

fgheysels:
This looks good to me.  
Maybe @gverstraete can have a look as well, since he has requested the feature ?

fgheysels (approved):

gverstraete:
Looks good to me :) Now just include it in the sas key validator :)

fgheysels:
> Looks good to me :) Now just include it in the sas key validator :)

Let's merge it then!

PR #365, opened 2022-09-07
fix: use correct namespace for versioned secret extensions

### LINKED ISSUE #364: Move versioned secret provider extensions to `Arcus.Security.Core` namespace
Currently, the versioned secret provider extensions are located in an `...Extensions` namespace while this should be in the general `Arcus.Security.Core` namespace for better discovery.

Add an additional set of extensions in the `Arcus.Security.Core` namespace that provides versioned secret provider extensions.

stijnmoreels:
Use the `Arcus.Security.Core` namespace so that consumers discover these extensions quicker.

Closes #364

### DIFF
--- a/src/Arcus.Security.Core/Extensions/ISecretProviderExtensions.Deprecated.cs
+++ b/src/Arcus.Security.Core/Extensions/ISecretProviderExtensions.Deprecated.cs
@@ -4,8 +4,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using GuardNet;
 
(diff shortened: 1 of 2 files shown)


fgheysels (approved):
To bad that we didn't include this with the 1.8.0 release, as this is now actually a small breaking change.

stijnmoreels:
> To bad that we didn't include this with the 1.8.0 release, as this is now actually a small breaking change.

Not necessarily, this PR adds deprecation on the released ones and adds new ones in the right namespace.

PR #373, opened 2022-11-24
feat: sync secret retrieval via secret store

### LINKED ISSUE #257: Provide the capability to add synchronous secret retrievals for certain secret provider instances
**Explain the problem/feature you want to have:**
Right now, all our `ISecretProvider` implementations and dependent interfaces uses this asynchronous paradigm to retrieve secrets. In some cases though, like User Secrets, Environment Variables, Docker Secrets..., where there's no external dependency involved, the secret can be retrieved synchronous. This is only 'hidden' behind the asynchronous wall (which is a good thing!).

In some cases though, like when registering the JWT authorization and you request the security token, or you want the instrumentation key of your Application Insights instance... you're stuck with a asynchronous system in a synchronous startup method. It would be cleaner if we introduce an additional interface for the `ISecretProvider` which synchronously retrieves the secret so when you want to get the instrumentation key, you could do:

```csharp
public void Configure(IApplicationBuilder app)
{
    var secretStore = app.ApplicationServices.GetRequiredService<ISecretStore>();
    var secretProvider = secretStore.GetProvider<ISecretProviderSync>("EnvironmentVariables");
    string instrumentationKey = secretProvider.GetRawSecret("APPLICATIONINSIGHTS_INSTRUMENTATIONKEY");
}
```

_Originally posted by @stijnmoreels in https://github.com/arcus-azure/arcus.security/discussions/254_

stijnmoreels:
Add support for retrieving secrets synchronously via the secret store, plus allow all built-in secret providers to benefit from this new functionality.

Closes #257

### DIFF
--- a/docs/preview/03-Features/secret-store/create-new-secret-provider.md
+++ b/docs/preview/03-Features/secret-store/create-new-secret-provider.md
@@ -165,46 +165,47 @@ namespace Application.Controllers
 }
 ```
 
-### Adding secret versions to your secret provider
-When your secret storage location supports versioned secrets, you could consider adapting your secret provider to support these.
-For more information on how you can use multiple versions of a secret in your application, see [this dedicated page](./versioned-secret-provider.md).
+For more information on consuming cached secrets, see [this dedicated page on consuming secrets](../secrets/general.md).
 
-Implement from `IVersionedSecretProvider` instead of `ISecretProvider` to allow the secret store to pick that your secret provider supports secret versions.
-The following example shows how the registry secret provider only supports two versions of a secret:
+### Adding synchronous secret retrieval to your secret provider
+When you want to use your secret provider in synchronous environments, you can implement the `ISyncSecretProvider` instead (implement also `ISecretProvider`).
+In our example, we can interact with the Registry synchronously, so we can benefit from synchronous secret retrieval.
 ```csharp
 using Arcus.Security.Core;
 using Microsoft.Win32;
 
-public class RegistrySecretProvider : IVersionedSecretProvider
+namespace Application.Security.CustomProviders
 {
-    // Also implement the general `ISecretProvider` methods...
+    public class RegistrySecretProvider : ISyncSecretProvider
+    {
+        public Task<string> GetRawSecretAsync(string secretName)
+        {
+            string secretValue = GetRawSecret(secretName);
+            return Task.FromResult(secretValue);
+        }
 
-     public Task<IEnumerable<string>> GetRawSecretsAsync(string secretName, int amountOfVersions)
-     {
-         if (amountOfVersions >= 2)
-         {
-             object valueV1 = Registry.LocalMachine.GetValue("v1\\" + secretName);
-             object valueV2 = Registry.LocalMachine.GetValue("v2\\" + secretName);
+        public Task<Secret> GetSecretAsync(string secretName)
+        {
+            string secretValue = GetRawSecret(secretName);
+            return Task.FromResult(new Secret(secretValue));
+        }
 
-             return Task.FromResult(new[] { valueV1, valueV2 });
-         }
+        public Secret GetSecret(string secretName)
+        {
+            string secretValue = GetRawSecret(secretName);
+            return new Secret(secretValue);
+        }
 
-         object valueV1 = Registry.LocalMachine.GetValue("v1\\" + secretName);
-         return Task.FromResult(new[] { valueV1 });
-     }
-
-     public async Task<IEnumerable<Secret>> GetSecretAsync(string secretName, int amountOfVersions)
-     {
-         string secretValue = await GetRawSecretAsync(secretName);
-         return new Secret(secretValue);
-     }
+        public string GetRawSecret(string secretName)
+        {
+            object value = Registry.LocalMachine.GetValue(secretName);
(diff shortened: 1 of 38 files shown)


docs/preview/03-Features/secret-store/create-new-secret-provider.md
```diff
@@ -165,46 +165,47 @@ namespace Application.Controllers
...
 
-### Adding secret versions to your secret provider
-When your secret storage location supports versioned secrets, you could consider adapting your secret provider to support these.
-For more information on how you can use multiple versions of a secret in your application, see [this dedicated page](./versioned-secret-provider.md).
+For more information on consuming cached secrets, see [this dedicated page on consuming secrets](../secrets/general.md).
 
-Implement from `IVersionedSecretProvider` instead of `ISecretProvider` to allow the secret store to pick that your secret provider supports secret versions.
-The following example shows how the registry secret provider only supports two versions of a secret:
+### Adding synchronous secret retrieval to your secret provider
+When you want to use your secret provider in synchronous environments, you can implement the `ISyncSecretProvider` instead (implement also `ISecretProvider`).
```
fgheysels, line 171: What is the reason why this is done with a separate interface ?
I don't see any benefit for that ?  Don't you think it becomes more complex to use this ? 

The issue here is, that you -as consumer- are obliged to know if you want to inject an ISecretProvider or an ISyncSecretProvider, and this breaks imho the abstraction.

docs/preview/03-Features/secrets/general.md
```diff
@@ -16,14 +16,41 @@ string secretVersion = secret.Version;
...
+```csharp
+var services = new ServiceCollection();
+services.AddSingleton(serviceProvider =>
+{
+    // #1 injecting the `ISyncSecretProvider`:
+    var syncSecretProvider = serviceProvider.GetRequiredService<ISyncSecretProvider>();
+    Secret secret = synSecretProvider.GetSecret("EventGrid-AuthKey");
+    string secretValue = syncSecretProvider.GetRawSecret("EventGrid-AuthKey");
+
+    // #2 calling `GetSecret` or `GetRawSecret` extension on `ISecretProvider`:
```
fgheysels, line 43: Ok, this extension method makes it easier to use (hence my earlier comment on 'why 2 different interfaces').
I think I now also understand why you decided to separate this functionality in another interface; it's because not all providers support this (like hashicorp).  However, we could've decided to in that case throw a `NotSupportedException`.
(Which is also probably what the extension method will do, if called on a `ISecretProvider` implementation that does not support sync secret retrieval?

docs/preview/03-Features/secrets/general.md
```diff
@@ -16,14 +16,41 @@ string secretVersion = secret.Version;
...
+    var secretProvider = serviceProvider.GetRequiredService<ISecretProvider>();
+    Secret secret = secretProvider.GetSecret("EventGrid-AuthKey");
+    string secretProvider = secretProvider.GetRawSecret("EventGrid-AuthKey");
+
+    return new MyDependentService(secretValue);
+});
+```
+
+⚠ Make sure that you only call the `GetSecret` and `GetRawSecret` extension on `ISecretProvider` implementations that also implement the `ISyncSecretProvider` interface. The [Arcus secret store](../secret-store/index.md) automatically makes sure that you can use this extension on any injected `ISecretProvider` but when no secret provider is registered that supports synchronous secret retrieval, an `SecretNotFoundException` will be thrown nonetheless.
+
```
fgheysels, line 53: Shouldn't we throw a `NotSupportedException` in that case.  I think it makes it more clear in that case of what is happening.   Throwing a  `SecretNotFoundException` might just confuse the user, as he's sure that there exists a secret with that specific key.

docs/preview/03-Features/secret-store/create-new-secret-provider.md
stijnmoreels: If we would alter the `ISecretProvider`, we would not only break contract, but we would also force any implementation to be synchronous. Our HashiCorp Vault, for example, cannot do that, and I imagine other implementations that are not able to do that.
Which is already the same conclusion you came to below this one 😉 .

docs/preview/03-Features/secrets/general.md
stijnmoreels: Yes, `SecretNotFoundException` with an `InvalidOperationException` as inner exception.

docs/preview/03-Features/secrets/general.md
stijnmoreels: There is an `InvalidOperationException` as an inner exception of the `SecretNotFoundException` to explain why the secret was not found. Wouldn't that be enough?

fgheysels: Idk if I would throw the `SecretNotFoundException` in this case, as it might be confusing.  It's actually not really the reason of the problem.

stijnmoreels: Ok, let's throw a `NotSupportedException` for this. 👍

fgheysels (approved):
Just spotted 2 'grammar errors'.
But, this is a nice feature, as it will allow you to setup dependencies in async methods, so less cumbersome code in asp.net projects :)

PR #389, opened 2022-12-26
chore: ability to push single package

stijnmoreels:
Add ability to push a single NuGet package with the YAML release pipeline by passing a package name as parameter.

### DIFF
--- a/build/nuget-release.yml
+++ b/build/nuget-release.yml
@@ -6,6 +6,9 @@ pr: none
 parameters:
   - name: 'Package.Version'
     type: 'string'
+  - name: 'Package.Name'
+    type: 'string'
+    default: '*'
 
 resources:
   repositories:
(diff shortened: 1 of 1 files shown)


stijnmoreels:
This was something you proposed, Frederik, but I have some questions regarding this.
If we allow a single package to be released, should we still release a GitHub release? This will also include a tag which makes me wonder if we release the same version of a previously (single-package) release, we would end up with the same release/tag.

build/nuget-release.yml
```diff
@@ -6,6 +6,9 @@ pr: none
 parameters:
   - name: 'Package.Version'
     type: 'string'
+  - name: 'Package.Name'
+    type: 'string'
+    default: '*'
```
fgheysels, line 11: In what cases do we want that ?

stijnmoreels: The cases you proposed 😄 . If you want to update a single package without the other packages to update.

fgheysels: To be honest, I can't remember that.
And since it seems we do not have an issue for this on GH and the fact that we would run into some other issues as you mention here, I think we should just abandon this idea.

PR #396, opened 2023-01-13
chore: use coverlet code coverage

stijnmoreels:
Use Coverlet code coverage as GitHub action.

Relates to https://github.com/arcus-azure/arcus/issues/264

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -35,12 +35,14 @@ jobs:
 
     - name: ReportGenerator
       uses: danielpalme/ReportGenerator-GitHub-Action@5.1.13
+      if: always()
       with:
         reports: 'src/Arcus.Security.Tests.Unit/lcov.info'
         targetdir: 'coveragereport'
 
     - name: Upload coverage report artifact
       uses: actions/upload-artifact@v2.2.3
+      if: always()
       with:
         name: CoverageReport
         path: coveragereport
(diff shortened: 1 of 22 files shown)


.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,45 @@
+name: "Code Coverage"
```
fgheysels, line 1: Since we have pipelines in Azure DevOps, I was thinking that the code-coverage report could be something that was part of the DevOps pipeline.
You could make sure that the tasks that run the test also generate the coverage statistics and then generate a report from them.

What exactly does this workflow do ?  It runs the tests yet again ?

stijnmoreels: Since all our other reports and analysis are done in GitHub/Actions (security analysis, dependency analysis, code style...), it seemed good to do our code-coverage also this way. We currently use Azure DevOps only for our CI/Release pipeline and as I understand it, it will also be (in the far future) be moved to GitHub. Another reason for me to already do this here. 
This workflow runs the Coverlet stuff after a test-run, that's correct. Like our CodeQL analysis also builds again. We will have a report here as artifact instead of going to Azure DevOps.

fgheysels: Ok, makes sense.

But then again, why don't we move our DevOps pipelines for CI builds and releases also to GH actions ?  That's a different discussion and has nothing to do with this PR, but I think we should consider that.

fgheysels (approved):
Good job extending the test-suite based on the coverage results.  (Although it could've been done in a separate PR).

stijnmoreels:
> Good job extending the test-suite based on the coverage results. (Although it could've been done in a separate PR).

Excluded the test functionality and the provider projects as this last one is mostly integration stuff towards other environments (i.e. Azure Key Vault, HashiCorp Vault...) and these are tested in integration tests.

stijnmoreels:
> Good job extending the test-suite based on the coverage results. (Although it could've been done in a separate PR).

Could be in a diff PR, but I added 80% as a threshold to be above, and wanted to be above that here. 👍

PR #398, opened 2023-01-16
chore: use codecov reporting

stijnmoreels:
Use Codecov reporting.


stijnmoreels:
What do you think of this kind of reporting? Using Codedev, we can also create a status badge afterwards with a percentage in our `README.md` file. Right now, because `main` has never run with code coverage, it only shows the percentage of the PR.

fgheysels:
Code coverage will be really helpful to see what code has not been tested, and if we can live with that.
But I can already see we're at a very high coverage percentage :)  Great job.

fgheysels (approved):

PR #427, opened 2024-01-22
chore: upgrade to observability v3.0

stijnmoreels:
Upgrade the projects to use Arcus Observability v3.0.

### DIFF
--- a/src/Arcus.Security.Tests.Integration/IntegrationTest.cs
+++ b/src/Arcus.Security.Tests.Integration/IntegrationTest.cs
@@ -27,7 +27,7 @@ namespace Arcus.Security.Tests.Integration
             var configuration = new LoggerConfiguration()
                 .WriteTo.XunitTestLogging(testOutput)
                 .WriteTo.Sink(InMemoryLogSink)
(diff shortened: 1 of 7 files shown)


src/Arcus.Security.Tests.Integration/IntegrationTest.cs
```diff
@@ -27,7 +27,7 @@ public IntegrationTest(ITestOutputHelper testOutput)
             var configuration = new LoggerConfiguration()
                 .WriteTo.XunitTestLogging(testOutput)
                 .WriteTo.Sink(InMemoryLogSink)
-                .WriteTo.AzureApplicationInsights(Configuration.GetValue<string>("Arcus:ApplicationInsights:InstrumentationKey"));
+                .WriteTo.AzureApplicationInsightsWithInstrumentationKey(Configuration.GetValue<string>("Arcus:ApplicationInsights:InstrumentationKey"));
```
fgheysels, line 30: I think it would be good if we could use the ConnectionString approach here instead of Instrumentation-Key.
Something for another PR ?

stijnmoreels: Yes, could be. 👍 Just upgraded the one with the same functionality here.
Created an issue for this: #428

fgheysels (approved):

PR #432, opened 2024-05-07
chore: upgrade az identity in keyvault secret provider

stijnmoreels:
Upgrade to first next version of `Azure.Idenity` that is not vulnerable in the Azure Key vault secret provider package.

### DIFF
--- a/src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
+++ b/src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
@@ -24,7 +24,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.10.2" />
+    <PackageReference Include="Azure.Identity" Version="1.11.0" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.5.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.0" />
     <PackageReference Include="Polly" Version="7.2.1" />


fgheysels (approved):

src/Arcus.Security.Providers.AzureKeyVault/Arcus.Security.Providers.AzureKeyVault.csproj
```diff
@@ -24,7 +24,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.10.2" />
+    <PackageReference Include="Azure.Identity" Version="1.11.0" />
```
fgheysels, line 27: Just wondering, why not the latest version (if this is not the latest version - which I assume because of the PR description).

fgheysels: 1.11.3 is the latest version it seems:
https://github.com/Azure/azure-sdk-for-net/releases/tag/Azure.Identity_1.11.3

stijnmoreels: Yes, but choose the 'minimum latest' as to not force the version upon people. We did this previously, too, so we are more broadly supported.

fgheysels: The latest version is a patch that contains some bugfixes .... I dont see a problem in using that latest version here.

stijnmoreels: It's more a library dev idea than a product dev idea. As in that we try to use the best yet lowest supported package. Since we recently released a major version, we could say that people should expect that there are some dependent packages updates, too, so we could indeed go forward with using v1.11.3, but otherwise, I believe that library development practices say that you should be as broadly supported; in this case highest version with vulnerability fix, lowest version that has that.

fgheysels (approved):
Should be updated again I think =D

Issue #434, opened 2024-08-23
Phase out the use of Guard.NET


fgheysels:
We'll remove the dependency Arcus.Security has on the Guard.NET package.
See [this](https://github.com/arcus-azure/arcus/discussions/362) discussion.

joachimgoris:
I can pick this one up

PR #435, opened 2024-09-17
chore: remove guard.net from core and keyvault secret provider

stijnmoreels:
Remove the `Guard.Net` package from the core and Azure Key vault secret provider projects.

Relates to #434


fgheysels (approved):
Less dependencies !

PR #436, opened 2024-09-27
chore: remove guard.net from other secret providers

stijnmoreels:
Remove the Guard.NET package references from the other secret provider projects.

Relates to #434

### DIFF
--- a/src/Arcus.Security.Providers.CommandLine/CommandLineSecretProvider.cs
+++ b/src/Arcus.Security.Providers.CommandLine/CommandLineSecretProvider.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Threading.Tasks;
 using Arcus.Security.Core;
-using GuardNet;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Configuration.CommandLine;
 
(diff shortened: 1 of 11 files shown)


src/Arcus.Security.Providers.CommandLine/CommandLineSecretProvider.cs
```diff
@@ -62,8 +56,6 @@ public Task<string> GetRawSecretAsync(string secretName)
         /// <exception cref="SecretNotFoundException">Thrown when the secret was not found, using the given name.</exception>
         public Secret GetSecret(string secretName)
         {
-            Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Requires a non-blank secret name to look up the command line argument secret");
```
fgheysels, line 65: we don't check for null here anymore because we rely on the null-check in `GetRawSecret` ?
Is that a good idea ?

stijnmoreels: Normally, no, but in this case the percentage of overloading functionality is greater than actual functionality. Since we do not need to use the values, only pass it along. It makes it  bit more easier to follow + only validates when it needs it.

Think it is discussible of what would be best. In a team with lots of inputs/members, I would indeed do paranoia-checks; but I might be bit pragmatic in this case.

Eager to hear your thoughts on this.

fgheysels (approved):

PR #441, opened 2025-05-02
chore(az-func): remove built-in az func support

stijnmoreels:
According to the minimization exercise described in #438 , (and like we did with Arcus.Messaging). We should limit ourselves to the core-business of the repositories and stop providing additional functionality for Azure Functions as it only hinders feature development.

We can remove the entire project, as it represents a separate NuGet package.


fgheysels (approved):
In fact, and if I'm correct, the Arcus.Security functionality can still be used in Azure Functions that use the isolated model (which should be the preferred model going forward), since the 'regular' Arcus.Security packages can still be used there ?

stijnmoreels:
> In fact, and if I'm correct, the Arcus.Security functionality can still be used in Azure Functions that use the isolated model (which should be the preferred model going forward), since the 'regular' Arcus.Security packages can still be used there ?

Yes, that's true.

PR #445, opened 2025-05-27
chore(core): deprecate 'get raw secret' exposure

stijnmoreels:
As discussed in #438 , the 'raw secret' functionality is something from the past and only adds unnecessary/duplicated maintainability overhead while not providing enough added-value.

This PR deprecates all mentions of the 'get raw secret' functionality in the core component.

### DIFF
--- a/docs/preview/03-Features/secret-store/versioned-secret-provider.md
+++ b/docs/preview/03-Features/secret-store/versioned-secret-provider.md
@@ -57,9 +57,6 @@ public class OrderController : ControllerBase
     {
         // Get all secrets available.
         IEnumeration<Secret> secrets = await _secretProvider.GetSecretsAsync("MySecret");
-
-        // Get all secret values available.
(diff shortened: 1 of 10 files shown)


src/Arcus.Security.Core/Caching/CachedSecretProvider.cs
```diff
@@ -89,6 +89,7 @@ public CachedSecretProvider(ISecretProvider secretProvider) :
         /// <exception cref="ArgumentException">The name must not be empty</exception>
         /// <exception cref="ArgumentNullException">The name must not be null</exception>
         /// <exception cref="SecretNotFoundException">The secret was not found, using the given name</exception>
+        [Obsolete("Will be removed in v3 in favor of solely using " + nameof(GetSecret) + " instead")]
```
fgheysels, line 92: When making these methods obsolete, shouldn't we (in the same PR?) also already introduce the implicit cast operator that allows to convert from `Secret` to `string` ?

PR #446, opened 2025-07-31
chore(net10.0): add support for .net 10.0 throughout

stijnmoreels:
Adds .NET 10 support throughout the solution mostly for more easily migration to .NET 10 without already removing Arcus in one go.


fgheysels:
Seems like some tests are using (now) deprecated functionality.

PR #455, opened 2025-08-19
chore(core): add new isecretprovider interface

### LINKED ISSUE #450: ✨ [New]: add new `ISecretProvider` interface in the `Arcus.Security` namespace
To prepare for the migration to a more manageable secret store (see #438), we should already create the necessary target functionality.

Add a new `ISecretProvider` interface in the `Arcus.Security` namespace in the **Arcus.Security.Core** project with the following members:

```csharp
public interface ISecretProvider
{
    SecretResult GetSecret(string secretName);
    Task<SecretResult> GetSecretAsync(string secretName);
}

public class SecretResult
{
    // Should have success/failure factory methods and expose the secret name/value/version/failure message/exception if applicable.
}
```

**Additional context**
See the #438 for more background information.

stijnmoreels:
Adds the proposal for a new `ISecretProvider` interface (currently `internal` to limit changes/warnings/errors).

Closes #450

### DIFF
--- a/src/Arcus.Security.Core/ISecretProvider.cs
+++ b/src/Arcus.Security.Core/ISecretProvider.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Threading.Tasks;
+using GuardNet;
 
 namespace Arcus.Security.Core
 {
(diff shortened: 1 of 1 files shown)


src/Arcus.Security.Core/ISecretProvider.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Threading.Tasks;
+using GuardNet;
```
stijnmoreels, line 3: **nore:** currently using `GuardNet` bc we still are supporting `netstandard2.1` in this PR. Will be removed in #446.

src/Arcus.Security.Core/ISecretProvider.cs
```diff
@@ -30,3 +31,212 @@ public interface ISecretProvider
...
+        /// </summary>
+        /// <param name="secretName">The name to identity the stored secret.</param>
+        /// <returns>
+        ///     <para>[Success] when the secret with the provided <paramref name="secretName"/> was found;</para>
+        ///     <para>[Failure] when the secret could not be retrieved via the provider.</para>
+        /// </returns>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="secretName"/> is blank.</exception>
+        Task<SecretResult> GetSecretAsync(string secretName)
+        {
+            return Task.FromResult(GetSecret(secretName));
```
fgheysels, line 53: Aren't we going to shoot ourselves in the foot with this default interface method ?
It looks like the 'Async' version for retrieving a secret won't be really 'async'.

stijnmoreels: Currently, a lot of our secret providers are solely 'sync' providers without an 'async' possibility: environment variables, configuration, Docker Secrets, User Secrets, Command Line... are all solely 'sync'. Having this in the interface greatly reduces the noise when implementing it (which would mean moving/duplicating the selected line to each and every one of those secret providers ➡️ which is currently done).

Issue #448, opened 2025-08-14
🏷️[Deprecate]: outdated functionality before v3


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Some functionality is already rather old and has increasing technical debt. We should make sure that we start building a newer secret store approach (see #438 ) that is prepared for the future.

**Describe the solution you'd like**
First step is to deprecate the following services:
* [ ] `CacheS

**Additional context**
See #438 for background inforation.

PR #456, opened 2025-08-21
chore(deprecate): secret caching via provider ifo secret store caching

stijnmoreels:
Deprecate all secret caching-related functionality in favor of moving the secret caching to the secret store.

Relates to #448


stijnmoreels:
Hmm, will split differently.

PR #460, opened 2025-08-22
chore(core): add secrets result proposal

stijnmoreels:
Versioned secrets will be moved to the concrete secret provider implementations instead of having it globally accessible via the secret store. We could, however, provide a common model that helps with this process: `SecretsResult` which is an aggregated `SecretResult` and helps when secret provider implementations want to return multiple secrets at once.

Relates to #438


stijnmoreels:
Will split differently.

PR #484, opened 2025-09-30
chore(core): make secret result name always available

stijnmoreels:
Make sure that all failed versions of the `SecretResult` model also includes the secret name for better tracability/defect localization. It also makes more sense that this information is always available.

Follow-up on #460

### DIFF
--- a/src/Arcus.Security.Core/ISecretProvider.cs
+++ b/src/Arcus.Security.Core/ISecretProvider.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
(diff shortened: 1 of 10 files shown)


src/Arcus.Security.Core/ISecretProvider.cs
```diff
@@ -200,7 +205,7 @@ public static SecretResult Interrupted(string failureMessage, Exception failureC
         /// <summary>
         /// Gets the secret value that was retrieved from the secret provider.
         /// </summary>
-        public string Name => IsSuccess ? _secretName : throw new InvalidOperationException($"[Arcus] cannot get secret name as the secret retrieval failed: {_failureMessage}", _failureCause);
+        public string Name { get; }
```
fgheysels, line 208: There's a bit of a discrepancy imho between this property-name (`Name`) and the parameter-names of the factory-methods (`secretName`).  Should this property also be named `SecretName`?

fgheysels: Indeed, I agree that it would add noise.
Renaming the parameters would be a better option then; it's not a must for me, I'll let you decide :)

fgheysels (approved):

